PR #154, opened 2024-03-03
improve portability

kavigupta:


create_website.py
```diff
@@ -296,31 +294,31 @@ def main(no_geo=False, no_data=False, no_juxta=False, no_data_jsons=False):
     if not no_juxta:
         generate_quiz_info_for_website("/home/kavi/temp/site")
```
lukebrody, line 295: Missed this one

kavigupta: oh whoops, see #155

PR #159, opened 2024-03-03
cleanup dependencies

lukebrody:
- remove unused/unnecessary react packages
- organize react packages into dev and not dev dependencies
- add python dependencies


react/package.json
```diff
@@ -17,23 +16,26 @@
...
+    "webpack": "^5.88.2",
     "@babel/core": "^7.22.9",
     "@babel/plugin-proposal-class-properties": "^7.18.6",
     "@babel/plugin-syntax-dynamic-import": "^7.8.3",
     "@babel/preset-env": "^7.20.2",
     "@babel/preset-react": "^7.18.6",
+    "babel-loader": "^9.1.0",
+    "downloads-folder": "^3.0.3"
+  },
+  "dependencies": {
```
kavigupta, line 33: Why are there two dependencies dicts

react/test/article_test.js
```diff
@@ -1,6 +1,6 @@
 import { Selector, ClientFunction } from 'testcafe';
 
-const TARGET = "http://teroknor:8000"
+const TARGET = "http://localhost:8000"
```
kavigupta, line 3: Making this an argument would be ideal

react/package.json
lukebrody: One is devDependencies
> "dependencies": Packages required by your application in production.
> "devDependencies": Packages that are only needed for local development and testing.

kavigupta: No like you have devdependencies and then two dependencies keys

kavigupta: Oh wait I misinterpreted the diff

react/test/article_test.js
lukebrody: added env var?

kavigupta (approved):

PR #205, opened 2024-05-13
"Northern" and "Southern"

glacialcascade:
Compare https://urbanstats.org/article.html?longname=Linyi+%28Northern+Shandong+Sheng%29+Urban+Center%2C+China and https://urbanstats.org/article.html?longname=Linyi+%28Southern+Shandong+Sheng%29+Urban+Center%2C+China

If lat2 > lat1, then 2 is more north than 1


kavigupta:
thanks for the catch and fix! I'll add a mention of you on the data credit page, should I cite you as glacialcascade or a twitter username?

PR #295, opened 2024-08-17
test for quiz that spins up a database instance and proxies it

kavigupta:
need to add tests for

- [x] not reporting scores that are older than the most recent score
- [x] retrostat
- [x] % correct numbers being added in
- [x] multiple different users having their results entered correctly
- [x] new user creation


react/test/quiz_test.js
```diff
@@ -1,14 +1,81 @@
...
+        .page(url)
+        // no local storage
+        .beforeEach(async t => {
+            // create a temporary file
+            const tempfile = "/tmp/quiz_test_" + Math.floor(Math.random() * 1000000) + ".sql";
+            // write the sql statements to the temporary file
+            writeFileSync(tempfile, sql_statements);
+            exec(`rm ../urbanstats-persistent-data/db.sqlite3; cd ../urbanstats-persistent-data; cat ${tempfile} | sqlite3 db.sqlite3; cd -`);
+            exec("bash ../urbanstats-persistent-data/run_for_test.sh");
+            await t.wait(2000);
```
lukebrody, line 58: 😭  
https://stackoverflow.com/a/56095793
```suggestion
            await exec(`rm ../urbanstats-persistent-data/db.sqlite3; cd ../urbanstats-persistent-data; cat ${tempfile} | sqlite3 db.sqlite3; cd -`);
            await exec("bash ../urbanstats-persistent-data/run_for_test.sh");
```

kavigupta: made the first change awaited; the issue is for the second that it needs to run async in the background

react/test/quiz_test.js
```diff
@@ -1,43 +1,284 @@
...
+function quiz_fixture(fix_name, url, new_localstorage, sql_statements) {
+    fixture(fix_name)
+        .page(url)
+        // no local storage
+        .beforeEach(async t => {
+            // create a temporary file
+            const tempfile = "/tmp/quiz_test_" + Math.floor(Math.random() * 1000000) + ".sql";
+            // write the sql statements to the temporary file
+            writeFileSync(tempfile, sql_statements);
+            await new Promise((resolve, reject) => {
```
lukebrody, line 67: Using `util.promisify` can eliminate this boilerplate.

urbanstats-persistent-data/run_for_test.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+cd $(dirname $0)
+rm -r venv
+virtualenv -p python3 venv
+source venv/bin/activate
+pip3 install -r requirements.txt
```
lukebrody, line 7: Maybe we only need to do this once rather than every test? Obviously we don't want to add a complex caching system, but maybe this is something we can do in the Dockerfile? (and maybe put the existing python libs in a venv)

This can be a separate MR (or not done at all until it starts really biting us)

kavigupta: We can do that, but I'm pretty sure these end up being cached anyway, it just copies them over

react/test/quiz_test.js
kavigupta: fixed

PR #321, opened 2024-09-03
Add lint rules (except naming)

lukebrody:
Should also fix the data credit page


kavigupta (approved):
LGTM. No pixel changes remaining

PR #326, opened 2024-09-07
Fix stats panel ascending

lukebrody:
Will merge this once #321 lands


kavigupta:
We should add a test to prevent a regression

lukebrody:
> We should add a test to prevent a regression; ideally assertion based not image based

@kavigupta Added!

lukebrody:
Need to add a lint check for test.only!!!

kavigupta (approved):
LGTM

PR #337, opened 2024-09-16
Lb/repro ci images

kavigupta:


lukebrody:
Replaced by #340 #341 #342 #343 #344

PR #339, opened 2024-09-17
loading class

kavigupta:

PR #340, opened 2024-09-17
Fix map loading during tests

lukebrody:
We need to be able to wait for the map to load during tests to capture the correct screenshot.

Add some state to the map so that it sets a CSS class when not loaded. Avoid a bug where the map reloads infinitely, by ignoring `loading` state updates when updating the map version.

Modify the test utils to wait for that CSS class to not be present before taking a screenshot.


lukebrody:
(follow up to #339)

PR #364, opened 2024-10-05
(3/3) Remove unecessary reloads, waits

lukebrody:
Now that pages are responsive, we don't have to reload them for them to update.

Update a few comparison screenshots since the map no longer repositions on reload.


kavigupta (approved):

lukebrody:
Recreated failure by adding this line 
image

because `UserId` component is in a race with generating the persistent id

PR #373, opened 2024-10-06
Abstract away background color

kavigupta:


react/src/components/comparison-panel.tsx
```diff
@@ -452,20 +453,7 @@ function insert_missing(rows: ArticleRow[][], idxs: number[][]): ArticleRow[][]
...
-            >
-                <span className="serif" style={{ fontSize: '15px', fontWeight: 500 }}>Zoom to:</span>
-                <div style={{ width: '0.25em' }} />
-                {this.zoom_button(-1, 'black', () => { this.zoom_to_all() })}
-                {this.props.longnames.map((longname, i) => {
-                    return this.zoom_button(i, this.props.colors[i], () => { this.zoom_to(longname) })
-                })}
-            </div>
-        )
+        return <ComparisonMapButtons map={this} />
```
lukebrody, line 456: This is a huge sin.

react/src/page_template/template.tsx
```diff
@@ -30,6 +32,9 @@ export function PageTemplate({
 }): ReactNode {
     const [hamburger_open, set_hamburger_open] = useState(false)
     const [screenshot_mode, set_screenshot_mode] = useState(false)
+    const colors = useColors()
+
+    document.body.style.backgroundColor = colors.background
```
lukebrody, line 37: This is probably fine but should really be a useEffect

react/src/components/comparison-panel.tsx
kavigupta: Wait why?

lukebrody: The whole structure of this is fucked because we're using class components.

We should refactor the Map to be an functional component, but until then stuff like this is going to keep popping up.

The problem with class components is that dependency management is much harder compared to functional components.

lukebrody: So no changes needed now, but this does me sad.

react/src/page_template/template.tsx
kavigupta: fixed

PR #383, opened 2024-10-10
Stat Checkbox Tree

### LINKED ISSUE #362: Replace checkboxes with an alternate system
Checkboxes are a very limited solution that can only reduce the complexity of the data by a quadratic amount. This is clearly an issue in two cases (1) large categories such as Industry/Occupation/Weather (2) categories where there's a lot of variant/Cartesian product statistics (eg 2020 population, 2010 population, 2010-2020 population change, and then the same 3 for pw density and it's variants, segregation, etc)

Ideally a solution to this problem should allow a user to seamlessly select a set of statistics that matches what they're looking for, without having to fiddle with checkboxes

A nice to have would be a way to isolate just a couple statistics for a screenshot or sharable link.

lukebrody:
Closes #362


react/src/components/comparison-panel.tsx
```diff
@@ -106,67 +108,104 @@ export function ComparisonPanel(props: { joined_string: string, universes: strin
...
+    const idxs: number[][] = []
+    const exclusively_american = props.datas.every(x => longname_is_exclusively_american(x.longname))
+    const settings = useSettings(groupYearKeys())
+    const statPaths = new Set<StatPath>()
+    for (const i of props.datas.keys()) {
+        const { result: [r, idx], availableStatPaths } = load_article(curr_universe, props.datas[i], settings,
+            exclusively_american)
+        rows.push(r)
+        idxs.push(idx)
+        availableStatPaths.forEach(path => statPaths.add(path)) // TODO: Maybe this should be an intersection?
```
lukebrody, line 122: @kavigupta

react/src/page_template/statistic-settings.ts
```diff
@@ -0,0 +1,324 @@
...
+}
+
+function useSelectedYears(): number[] {
+    const availableYears = useAvailableYears()
+    const settingsValues = useSettings(yearKeys(availableYears))
+    return availableYears.filter(year => settingsValues[`show_stat_year_${year}`])
+}
+
+/**
+ * Which groups are selected, but are not showing any statistics because no years are selected.
```
lukebrody, line 226:
```suggestion
 * Which groups are selected, but are not showing any statistics because no relevant years are selected.
```

react/src/page_template/statistic-settings.ts
```diff
@@ -0,0 +1,324 @@
...
+ */
+export function useGroupsMissingYearSelection(): (Group | Category)[] {
```
lukebrody, line 228: TODO: Remove this

react/src/components/StatsTree.tsx
```diff
@@ -0,0 +1,108 @@
...
+    return useAvailableCategories().map(category => <CategoryComponent key={category.id} category={category} />)
+}
+
+function CategoryComponent({ category }: { category: Category }): ReactNode {
+    const categoryStatus = useCategoryStatus(category)
+    const [isExpanded, setIsExpanded] = useSetting(`stat_category_expanded_${category.id}`)
+    const isMobileLayout = useMobileLayout()
+    const changeCategorySetting = useChangeCategorySetting(category)
+    return (
+        // Move the category left half of the indent since it's not really a child
```
lukebrody, line 19:
```suggestion
```

PR #388, opened 2024-10-10
hue colors for quiz

kavigupta:
Seems like the new behavior is correct? (The new behavior shows up on urbanstats when I do it manually)

Before:

https://github.com/user-attachments/assets/f118c042-4d2a-4182-b250-98833c00d803

After: 

https://github.com/user-attachments/assets/71643271-2761-4b7a-916b-adc15916c604


kavigupta:
Made redundant by other commits

Issue #315, opened 2024-09-02
Ordering on histogram series is inaccurate in some situations


kavigupta:
In cases like USA vs United Kingdom it seems to be inaccurate; I think this is due to ASCII based vs alphabetical sorting: https://urbanstats.org/comparison.html?longnames=%5B%22USA%22%2C%22United+Kingdom%22%5D

I think a proper fix involves using the domain field: https://talk.observablehq.com/t/observable-plot-how-to-set-custom-colors-based-of-field-values/8095/3

PR #397, opened 2024-10-15
Hide pointer buttons in screenshots (#315 2/2)

lukebrody:
Before 

Delran township  CCD , Burlington County, New Jersey, USA

After

Delran township  CCD , Burlington County, New Jersey, USA (2)


lukebrody:
Also hide whitespace changes for this one

PR #443, opened 2024-10-21
Allow linking settings for articles

lukebrody:


lukebrody:
Existing tests pass 🙌

react/src/page_template/settings.ts
```diff
@@ -75,19 +76,37 @@ const defaultSettings = {
...
 
 export class Settings {
+    /**
+     * Basic Settings
+     */
     private readonly settings: SettingsDictionary
 
     constructor() {
-        this.settings = load_settings()
+        const savedSettings = localStorage.getItem('settings')
```
kavigupta, line 91: I'm not sure if we want new users to just have their settings set to whatever link they were sent

react/src/page_template/settings.ts
```diff
@@ -96,35 +115,151 @@ export class Settings {
...
 
     // Singular settings means we can use observers
     static Context = createContext(new Settings())
+
+    /**
+     * Staged Mode
+     */
+    private stagedSettings?: Partial<SettingsDictionary>
+
+    /**
```
kavigupta, line 161: Comment should explain what staging is

react/src/page_template/settings-vector.ts
```diff
@@ -0,0 +1,319 @@
...
+}
+
+sanityCheckVector()
+
+export type VectorSettingKey = typeof settingsVector[number]
+
+export function useVector(): string {
+    const settings = useSettings(settingsVector)
+    const booleans = settingsVector.map(setting => settings[setting])
+    return base58.binary_to_base58(compressBooleans(booleans))
```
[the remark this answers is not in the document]
lukebrody: base64 is ugly in URL params

react/src/page_template/settings.ts
lukebrody: Should we just show them the staging UI? Might be a lot for a new user. Plus if they stick around they were obviously interested in what they were sent

react/src/page_template/settings.ts
lukebrody: Added an explanation, lmk if you have a better one

react/src/page_template/settings.ts
kavigupta: Yeah maybe this is fine, wasn't considering cognitive load of the staging view

react/src/page_template/settings.ts
kavigupta: LG

react/src/page_template/settings.ts
lukebrody: and it won't get saved unless they check a box

lukebrody: I'll actually formalize this requirement in the tests

kavigupta (changes requested):
Apply then refresh does not work (it puts it back into staging). Somehow, apply isn't applying settings.

react/src/page_template/settings.ts
lukebrody: hmm ok how about this

we save another localstorage that's a set of settings that have been user modified

don't load settings from link, enter stage mode like normal

if the set of user modified settings does not intersect with the set of settings we're staging, we hide that we're in staging mode in the UI

if the user changes one of the keys that we're staging in this hidden mode, immediately apply the stage

this avoids trivial visual modifications (e.g. opening a category) from marking a user as having settings

lukebrody:
> Apply then refresh does not work (it puts it back into staging). Somehow, apply isn't applying settings.

Oof can confirm it doesn't look like apply is saving. Also should test this

lukebrody:
> Apply then refresh does not work (it puts it back into staging). Somehow, apply isn't applying settings.

Fixed this! Literally just forgot to save on apply

kavigupta (approved):

react/src/page_template/settings.ts
lukebrody: Going to not do the above now ^

PR #447, opened 2024-10-23
add symlink framework

kavigupta:
The idea here is to allow for renaming longnames. To preserve old URLs, we add "symlinks" written into a dictionary that map one to the other

This change is added to make the country renaming change work, and is completely adequate for that change (which only changes 1983 names) but might not scale well in future

## Alternative Considered

Change the `Article` data model in the protobuf to a Union type where one of the union possibilities is a redirect:

```
message ArticleContents {
    string shortname = 1;
    string longname = 2;
    string source = 3;
    string article_type = 4;
    repeated StatisticRow rows = 5;
    repeated RelatedButtons related = 6;
    repeated string universes = 7;
    repeated ExtraStatistic extra_stats = 8;
}

message ArticleRedirect {
    string actual_longname = 1
}

message Article {
    optional ArticleContents contentsn = 1
    optional ArticleRedirect redirect = 2
}
```

This alternative was rejected because it would require more requests on the frontend and slightly more complicated code. It, however, scales better


react/src/data/symlinks.json
```diff
@@ -0,0 +1 @@
+{"United States of America": "USA"}
```
lukebrody, line 1: How is this file updated?

kavigupta: it's updated in `create_website.py`

react/src/utils/symlinks.ts
```diff
@@ -0,0 +1,18 @@
+const SYMLINKS = require('../data/symlinks.json') as Record<string, string>
+
+export function followSymlink(name: string): [string, boolean] {
```
lukebrody, line 3:
```suggestion
export function followSymlink(name: string): string | undefined {
```
Union types pls

react/src/data/symlinks.json
lukebrody: Missed this

PR #450, opened 2024-10-23
Pipeline refactor/Better control when the docker image is built

lukebrody:


lukebrody:
This is too complicated 
#452 is better

PR #469, opened 2024-10-26
Lint for python

kavigupta:


.github/workflows/check.yml
```diff
@@ -17,6 +17,24 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
```
lukebrody, line 31: Just use the docker image!

probably need to add the requirements.txt to the image name

kavigupta: is that correct?

lukebrody: Yeah LGTM!

Issue #455, opened 2024-10-23
More general framework for universe-specific statistics and universe defaults


kavigupta:
Currently universe specific statistics and defaults are handled via checking to see if a region is "American" by looking for "USA" in the region's longname. This is both janky and will not scale well to having statistics that are, e.g., Canada specific

Issues that need to be resolved

- [ ] index lists; more generally the problem of figuring out which statistics go into a particular Article model and how to represent which ones are available
- [x] default universe; how to pick a default universe for a given article or comparison

PR #494, opened 2024-10-27
try to make the universe framework less US-centric

kavigupta:
Currently this is just a bunch of FIXME tags. We'll probably want to solve several of these issues separately. Will work towards resolving https://github.com/kavigupta/urbanstats/issues/455


kavigupta:
All fixmes resolved, closing

PR #516, opened 2024-11-01
Improve docker performance

lukebrody:
Decrease container size by 25%

Shaves about 10 seconds off container pull time

[docker optimization.pdf](https://github.com/user-attachments/files/17601681/docker.optimization.pdf)

Removes video recording from tests


lukebrody:
Saves 30s in the pipeline 🙌

Issue #506, opened 2024-10-31
Discord is still breaking links


lukebrody:
Broke this one

https://urbanstats.org/statistic.html?statname=Binge+drinking+among+adults+%%26article_type=Subnational+Region&start=1&amount=100&order=ascending&universe=USA

kavigupta:
```
Original error-causing page:
https://urbanstats.org/statistic.html?statname=Households+With+no+Vehicle+%25&article_type=Subnational+Region&start=21&amount=20&order=ascending&universe=USA
Original error caused:
https://urbanstats.org/statistic.html?statname=Households+With+no+Vehicle+%25%26article_type%3DSubnational+Region&start=21&amount=20&order=ascending&universe=USA
New error caused:
https://urbanstats.org/statistic.html?statname=Households+With+no+Vehicle+%25%26article_type%3DSubnational+Region&start=21&amount=20&order=ascending&universe=USA
```

kavigupta:
Ok I think this is just unfixable

PR #518, opened 2024-11-02
fix the discord links issue by just getting rid of % in links

kavigupta:
Resolves https://github.com/kavigupta/urbanstats/issues/506

### DIFF
--- a/react/src/navigation/links.ts
+++ b/react/src/navigation/links.ts
@@ -78,7 +78,7 @@ export function statistic_link(universe: string | undefined, statname: string, a
         start = start + 1
     }
     const params = new URLSearchParams()
-    params.set('statname', statname)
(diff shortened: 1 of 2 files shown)


lukebrody:
Won't this break some existing links? Also we should remove the existing patch probably?

lukebrody:
> Won't this break some existing links?

Nevermind I see we go the other way

kavigupta:
Yeah and the existing fix doesn't like, cause issues. We can remove it if we like

lukebrody:
> Yeah and the existing fix doesn't like, cause issues. We can remove it if we like

I guess it might break some links that are working if we remove it...

lukebrody:
might have been good to leave some comments around 😅

PR #529, opened 2024-11-03
More typed react data

lukebrody:


urbanstats/utils.py
```diff
@@ -43,3 +45,11 @@ def compute_bins(data, weight, *, bin_size=0.1):
     idx = np.clip(idx, 0, len(values) - 1)
     np.add.at(values, idx, weight)
     return values
+
+
+def output_typescript(data, file, dataType="const"):
```
lukebrody, line 50: This should probably be data_type

PR #546, opened 2024-11-09
make it so setting deprecation does not break the types

kavigupta:


react/src/page_template/settings-vector.ts
```diff
@@ -252,20 +252,21 @@ export function useVector(): string {
 
 export function fromVector(vector: string, settings: Settings): Record<VectorSettingKey, boolean> {
     const array = decompressBooleans(base58.base58_to_binary(vector))
-    return Object.fromEntries(settingsVector.map((setting, i) => {
+    const result = settingsVector.map((setting, i) => {
```
lukebrody, line 255:
```suggestion
    const result = settingsVector.flatMap((setting, i) => {
```

react/src/page_template/settings-vector.ts
```diff
@@ -252,20 +252,21 @@ export function useVector(): string {
...
             value = array[i]
         }
         else {
             value = settings.get(setting.key)
         }
-        return [setting.key, value]
-    }))
+        return [[setting.key, value]] satisfies [VectorSettingKey, boolean][]
+    })
+    return Object.fromEntries(result.flat())
```
lukebrody, line 269:
```suggestion
    return Object.fromEntries(result)
```

lukebrody:
LGTM

Issue #569, opened 2024-11-11
Avoid needing to update test links when adding new settings


lukebrody:
<img width="760" alt="image" src="https://github.com/user-attachments/assets/7a8b6b4c-22d7-4af1-a0c7-d1c83bff5a94">

lukebrody:
Fixed in #580

PR #573, opened 2024-11-12
Add the histogram type setting to the settings vector

### LINKED ISSUE #522: Add histograms to settings link
we can include those settings by appending to settingsVector and then including it in ArticleComparisonQuerySettingsConnection (if it also shows up in comparison, otherwise we should exclude it from comparison else irrelevant settings will trigger staging)
I guess I just overlooked this
we'll also need to have unchanging concrete forms of any settings we include in setttingsVector

Also should highlight the histogram in some way when staged

lukebrody:
Closes #522


lukebrody:
Need to add a test

lukebrody:
Will use this as a testbed for #569

PR #606, opened 2024-11-17
Fix the quiz test screaming

lukebrody:


kavigupta:
Hate it when my quiz test scream

PR #615, opened 2024-11-17
SPA

### LINKED ISSUE #668: Convert site to SPA
There will be one big change to switch to SPA
And then better progress indicators/dynamic loading (although the bundle is now only a few mb) will be smaller changes

Should also follow this up by fixing circular imports

lukebrody:
Closes #668 

Converts the site from a multi-page application to a single page application.

- Convert multiple entrypoints into one entrypoint. The one entrypoint parses the URL and loads the correct page.
- We can still support initial navigation to multiple pages by simply copying this one `index.html` entrypoint to the other page names.
- Introduce the `Navigator` abstraction, which handles navigation across the SPA.
  - Handles universes, settings vector, etc. These were previously ad-hoc which could make their interactions confusing to understand. Remove other contexts.
  - Handles most interaction with `location`, `history` APIs. Add eslint rules to ban use of these APIs generally.
- We should no longer use traditional links for navigation within the site. Instead, use the properties from `navigator.link`. Add an eslint rule to enforce this.
- Many modal UI elements (search, universe selector, add comparison article) should close after the navigation occurs, since a new page is not being loaded.
- The mobile sidebar should close once new navigations are complete, as a new page is no longer being loaded.
- Move setting the document title to be an effect of page panels rather than happening during loading.
- Various other fixes and refactors.
- Add some new tests for the navigator.


lukebrody:
My brain after doing this one: https://www.youtube.com/watch?v=4mo1nydqGxQ

lukebrody:
Need to add tests for navigation errors

lukebrody:
@kavigupta however much manual/review testing you want to do is up to you, I feel the code quality is pretty good, out test suite is quite comprehensive 
Landing this will definitely introduce some unknown bugs, but given how much testing we have, I feel they’re unlikely to be major.

PR #638, opened 2024-11-20
Fix setting vector types to ensure that setting types are respected

lukebrody:


react/src/page_template/settings-vector.ts
```diff
@@ -71,294 +71,295 @@ type Min2<T extends readonly unknown[]> = [...T] extends [unknown, infer X, ...i
...
+            return array[number] ?? underflow
+        },
     }
 }
 
-const HistogramTypeSettingCoder = new BitmapCoder(2, [
+const HistogramTypeSettingCoder = BitmapCoder(2, [
     'Line',
     'Line (cumulative)',
     'Bar',
```
lukebrody, line 110: Previously you could add or remove options here and no error would be thrown.

lukebrody:
Replaced by #639

PR #640, opened 2024-11-20
more gridded statistics

kavigupta:


kavigupta:
Hmm I'm thinking I can just disagg to blocks for localities and elevation

PR #653, opened 2024-11-22
Install chrome the recommended way

### LINKED ISSUE #649: Should use the recommended way to install chromium for testing
https://www.chromium.org/getting-involved/download-chromium/

This version is better than the consumer version

lukebrody:
Cuts chrome size in docker image from 693MB to 383MB

Remove browser choice since we're not using chromium anymore

Closes #649


lukebrody:
It's really slow and doesn't work

PR #689, opened 2024-12-06
Improve loading UX on slow connections

### LINKED ISSUE #495: improve loading UX over poor connections
On poor connection, urbanstats shows a blank screen with no loading indicator for some time

Address the root cause of long loading as much as possible, then improve the experience during loading

lukebrody:
Closes #495 

Before

https://github.com/user-attachments/assets/1016a470-1052-4990-bcb2-ea7464e8a30c

After 

https://github.com/user-attachments/assets/4cd6012c-a950-4cd9-882d-36738d6c1a54


lukebrody:
Hmm.... maybe want to have the spinner be just on index.html, and then keep page loading its own thing?

PR #729, opened 2024-12-11
Improve SPA

### LINKED ISSUE #715: Improve SPA navigation
SPA navigation exhibits the following problems:
- Sometimes, navigating between different types of pages retains the scroll position, even though the page should reset to the top.
- Going back/forward does not preserve scroll position.

Correct these problems by:
- Centralizing `popstate` management into the navigator.
- Having the navigator emit events when it finishes navigation, in order for effects to be applied, if necessary.
  - Move hash handing here.
  - Move scrollPosition handling here.
- Include the scrollPosition in the navigation state.

### LINKED ISSUE #716: SPA: Error: Already in staged mode
SPA throws this error when you're in staged mode and you navigate to a page that would cause entering staged mode again.

Instead, we should
- When you navigate to a descriptor that would put the settings in staging mode, and we're already in staging mode, stay in staging mode with the new settings.
- When you navigate to a descriptor without settings, exit staging mode (discard)

lukebrody:
- Get out of staging mode if we go to a non-settings-vector page.
- Changes to scroll position handling
  - Save scroll position as part of the history state
  - Restore scroll position on back/forward
  - Scroll to top when navigating to different kinds of pages
- Add somewhat graceful handling of history state evolution, use zod schemas to parse it, and reload if failure.
- Add an effect system for navigator page loads, so that loading a page can have effects that are executed after it renders. (This is used for scrolling ,etc)
- Move around some event listeners into the navigator rather than in the router.

Closes #715 
Closes #716


lukebrody:
Need to split this up, but just testing everything together for now

PR #730, opened 2024-12-11
juxtastat friends

### LINKED ISSUE #723: Juxtastat friends feature
## Concept

User interface: there's some "add friend" box where you can put in a friend's nickname and id. Nickname is stored only locally so no privacy risk. Whenever you finish a Juxtastat it tells you how well your friends did on that problem if they've played it at the bottom of the screen. Maybe it also lets you compare statistics but that might be too much complexity?

## Implementation

Persistent server: add an endpoint allowing access to a list of users' statistics for a given day, given their IDs.

Frontend: section at the bottom showing your friends scores on today's juxtastat. Button at the bottom letting you add to this. Whenever rendering the results page, send a request to the server for your friends scores.

```
/initate_friend_request
param user: the requester, who initates the request
param requestee: the requestee, who will receive the request

/unfriend
same thing as above, but removes the friend relation

/active_friend_requests
param user: the requestee, who initates the request
returns: the requesters who have made requests of user

/todays_score_for
param user: the requestee. For this to work must be the requestee for each requester in friends
param friends: the friends, to gather information on
returns: scores for each friend, or a special `not a friend` message for anyone who has not requested `user` as a friend

/poke
param user: the requestee. Same rules as todays_score_for
param friend: the requester

/see_pokes
param user: the requester
returns: a list of users who have initiated a poke and the server time of each poke, along with current server time
```

## Security considerations

this allows users to access arbitrary other users' Juxtastat scores, and, more importantly, when they do the Juxtastat with nothing but their persistent id, which is shown on screen in screenshots, etc. To solve this we should probably make this feature bidirectional, not giving Alice access to Bob's scores unless Bob has also requested access to Alice's. The issue with this is that it is trivial to impersonate a user currently in Juxtastat, since we only use persistent id to set a key. We should first roll out a change that authenticates users with a secret, non-displayed ID that is also stored in local storage and downloaded in persistent data dumps. This id can be used to authenticate requests and will not be accessible by anyone other than the user in question.

kavigupta:
Resolves #723 

- [x] make friend names editable
- [x] make this work for retrostat
- [x] add test for seeing a friends' scores
- [x] add test for having a friend be pending
- [x] add test for a friend not having done the quiz yet
- [x] add test for multiple friends
- [x] add test to make sure you can't have the same friend twice
- [x] add test to make sure unfriending works (you show up as pending again on the other friends' page)
- [x] add test for retrostat
- [x] add test for renaming a friend
- [x] alignment issue on add friend


lukebrody:
Can we be... 👉 👈  Juxtastat friends?

PR #741, opened 2024-12-12
Try to catch a flaky test

lukebrody:


lukebrody:
I think #742 will actually fix this flakiness, so closing this unless that's not he case

PR #746, opened 2024-12-13
Use camelCase only in react

lukebrody:
Pure rename, should be no functionality changes.

Add eslint rules for enforcement.


lukebrody:
Waiting on https://github.com/kavigupta/urbanstats/pull/730

PR #749, opened 2024-12-14
add friend tables and query system

kavigupta:
This should be merged and deployed before #730 is. It doesn't do anything without the frontend


lukebrody:
FRIEND TABLES

PR #766, opened 2024-12-17
fix quiz share link

### LINKED ISSUE #765: Juxtastat share link has quiz.html
Should just be juxtastat.org

lukebrody:
closes #765


kavigupta (changes requested):
Let's add a regression test every time we make a change like this

lukebrody:
> Let's add a regression test every time we make a change like this

Yeah good point, I thought this would be annoying to test but actually I think we can mock out navigator.copyText

lukebrody:
No kill like overkill

PR #773, opened 2024-12-18
Fix articles with many /

### LINKED ISSUE #759: SPA broke link sanitization
I think this bug was introduced by SPA. Following tests fail:

1. Type into the address bar "victory manor" then hit enter
2. Go to https://urbanstats.org/article.html?longname=Savannah+city%2C+Georgia%2C+USA&universe=North+America&s=FGeQWrY52z2nuF and then click on the button containing "victory manor". 
3. Go to https://urbanstats.org/statistic.html?statname=Population&article_type=Neighborhood&start=10341&amount=20&universe=North+America and click the link containing "victory manor".

I'm sure there's others I've missed. This can probably be fixed in fetching data links

lukebrody:
Closes #759 

This was always broken

### DIFF
--- a/react/src/navigation/navigator.tsx
+++ b/react/src/navigation/navigator.tsx
@@ -24,7 +24,7 @@ import { Article, IDataList } from '../utils/protos'
 import { followSymlink, followSymlinks } from '../utils/symlinks'
 import { NormalizeProto } from '../utils/types'
 
-import { dataLink, sanitize } from './links'
+import { dataLink } from './links'
 import { byPopulation, uniform } from './random'
 
 const articleSchema = z.object({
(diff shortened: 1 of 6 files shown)


react/src/navigation/navigator.tsx
```diff
@@ -187,15 +187,15 @@ export function urlFromPageDescriptor(pageDescriptor: ExceptionalPageDescriptor)
         case 'article':
             pathname = '/article.html'
             searchParams = {
-                longname: sanitize(pageDescriptor.longname),
```
lukebrody, line 190: @kavigupta This is going to break some recently shared links (it was a mistake in the implementation). Only those that need to be sanitized though, which should be few

kavigupta: Yeah that's fine

PR #790, opened 2024-12-23
custom quiz

kavigupta:


react/src/navigation/navigator.tsx
```diff
@@ -410,6 +413,15 @@ async function loadPageDescriptor(newDescriptor: PageDescriptor, settings: Setti
             let quizDescriptor: QuizDescriptor
             let todayName: string
             switch (newDescriptor.mode) {
+                case 'custom':
+                    const custom = JSON.parse(base64Gunzip(newDescriptor.quizContent ?? '')) as CustomQuizContent
```
lukebrody, line 417: Please use zod schemas here for better error handling

kavigupta: is there a way to do this in a way that doesn't require rewriting this whole thing into zod types?
image

lukebrody: .... no... probably not worth it I guess for this use case

PR #879, opened 2025-01-19
better tests for juxta infinite

kavigupta:


kavigupta:
Just realized this doesn't make any sense because it requires saving statistics

PR #900, opened 2025-02-01
put errors loading page into the console, for easier debugging

kavigupta:


lukebrody:
this is the wrong place to put this

kavigupta:
> Should go here
> 
> https://github.com/kavigupta/urbanstats/blob/master/react/src/navigation/Navigator.ts#L228
> 
> and here
> 
> https://github.com/kavigupta/urbanstats/blob/master/react/src/navigation/Navigator.ts#L73

fixed

PR #901, opened 2025-02-01
clicking copy twice should not crash

### LINKED ISSUE #902: clicking copy twice should not crash, but does

kavigupta:
Will resolve #902


lukebrody:
it's clicking the Copy Link button in the friends panel which asks to you to enter your name
we should use withExactText

Issue #830, opened 2024-12-29
`counts_by_article_type` as protobuf


lukebrody:
- Speeds up builds
- Probably smaller loading, doesn't have to parse as JS or be in the bundle
- Can load as required for pages in Navigator

PR #921, opened 2025-02-06
counts_by_article_type as a protobuf

kavigupta:
Resolves https://github.com/kavigupta/urbanstats/issues/830


react/src/navigation/PageDescriptor.ts
```diff
@@ -407,6 +409,7 @@ export async function loadPageDescriptor(newDescriptor: PageDescriptor, settings
                     data: await data,
                     renderedStatname: newDescriptor.statname,
                     universe: statUniverse,
+                    counts: await counts(),
```
lukebrody, line 412: This and the `await counts()` above it might fire off two calls.... either confirm the browser is smart enough to cache & combine them, or prefer one `const countsResult = await counts()` above. (Naming should probably be `const counts = await getCounts()`

kavigupta: Wait it's cached in the ts though right?

kavigupta: image

lukebrody: Oh missed this. In this case prefer making `loadArticle` async using `getCountsByArticleType()` within that function. This reduces complexity by moving code that deals with article type counts closer to where we need to care about it.

lukebrody: For this line you can add a simple `await getCountsByArticleType()` rather than needing a separate variable.

kavigupta: StatisticPanel needs this to compute the set of universes to display, I've added a comment to this effect

kavigupta: fixed the redundancy of `counts` as a variable

PR #980, opened 2025-02-19
attach type to search element

kavigupta:
to be used later on in adding icons to the search


kavigupta:
@lukebrody added you to review because I'm not sure if this is going to make it way slower or soething

PR #992, opened 2025-03-07
Search cache

lukebrody:
image


lukebrody:
with simulated latency (Slow 4g)

cache hit

image

cache miss

image

So actually we should return the index before we know if it's up-to-date... and then replace it if we can get a newer one

PR #993, opened 2025-03-07
Eliminate search key wait time by getting the key immediately

lukebrody:
This makes more sense anyhow as we're concerned with the version matching the loaded page


lukebrody:
All on slow 4g, fresh cache

Cache miss

image

Cache hit

image

PR #1174, opened 2025-06-16
Add fullscreen control to maps

### LINKED ISSUE #1170: Allow maps to be made full screen
https://maplibre.org/maplibre-gl-js/docs/examples/fullscreen/ - doesn't seem to be too difficult?

glacialcascade:
Adds a full screen button to all maps (in the top left so that it doesn't interfere with the "Zoom to" control in compare, though it might seem a bit odd that those have different heights/margins). Maybe worth asking if this is necessary for the small juxtastat maps but I don't think it's too cluttery.

Resolves #1170.

I also had to explicitly tell rspack not to watch node_modules in order to stay under the file watchers limit, though apparently this should be default behavior.


lukebrody:
@kavigupta Do we want this on quizzes? I could go either way.

quiz-clickthrough-test-1

react/rspack.config.js
```diff
@@ -76,5 +76,8 @@ export default env => ({
             }),
             new rspack.LightningCssMinimizerRspackPlugin(),
         ],
-    }
+    },
+    watchOptions: {
+        ignored: /node_modules/,
+    },
```
lukebrody, line 82: @kavigupta How annoying is this for Linux contributors? Watching libraries is nice for me, but there aren't any downsides on my platform.

lukebrody:
@glacialcascade Agree that the alignment with "Zoom To" could be better. Please try to fix that in this PR.

react/rspack.config.js
lukebrody: Sounds good, we can go ahead and upgrade Rspack in a separate PR rather than making this change here.

react/src/common.css
```diff
@@ -120,3 +120,27 @@ input[type=text] {
...
+        background-color: var(--text-main);
+        background-image: none;
+    }
+
+    > button.maplibregl-ctrl-shrink .maplibregl-ctrl-icon {
+        mask-image: url("data:image/svg+xml;charset=utf-8,%3Csvg xmlns='http://www.w3.org/2000/svg' width='29' height='29' viewBox='0 0 29 29'%3E%3Cpath d='M18.5 16c-1.75 0-2.5.75-2.5 2.5V24h1l1.5-3 5.5 4 1-1-4-5.5 3-1.5v-1zM13 18.5c0-1.75-.75-2.5-2.5-2.5H5v1l3 1.5L4 24l1 1 5.5-4 1.5 3h1zm3-8c0 1.75.75 2.5 2.5 2.5H24v-1l-3-1.5L25 5l-1-1-5.5 4L17 5h-1zM10.5 13c1.75 0 2.5-.75 2.5-2.5V5h-1l-1.5 3L5 4 4 5l4 5.5L5 12v1z'/%3E%3C/svg%3E");
+        background-color: var(--text-main);
+        background-image: none;
+    }
+}
```
lukebrody, line 146: Looks great! Let's go ahead and put this in a separate `map.css`, then import that into `map.tsx`

lukebrody:
Regarding the failing jobs:

- Download the last artifact generated by the `merge_screenshots` job.
- Review the changes under "delta".
- If the changes look as expected, copy the changes under changed_screenshots into urbanstats/reference_test_screenshots, and commit them to this branch.
- The test jobs should then succeed.

react/rspack.config.js
lukebrody: https://github.com/kavigupta/urbanstats/pull/1177

lukebrody:
You should pull in master and update the screenshots again to resolve conflicts.

lukebrody:
> The first time the tests were run after merging in master, `renders transpose comparision-1.png` was screenshotted while the page was still loading, and `search-test-arrows-1.png` did not succesfully select the 3rd search result (perhaps since it took too long to load the results?) - I have noticed that the search seems slower than usual recently

Hmm... maybe the runner was having a bad day... If we see this more often we can look into fixing it.

lukebrody:
>  I have noticed that the search seems slower than usual recently

If you turn on this flag and watch the console, you might be able to give us some insight on what's slow on your device.

https://github.com/glacialcascade/urbanstats/blob/b3c4d5ff9f4183c4edeeb665ce5d755c52b943ec/react/src/search.ts#L24

PR #1196, opened 2025-07-04
separate out checkbox

kavigupta:


kavigupta:
I believe it fixed it

kavigupta:
it seems i did not in fact fix it

kavigupta:
hmm not sure why this is tripping memory tests??

lukebrody:
It may be that there are so many checkboxes on California that making another React component does legitimately increase the memory usage. Not by much though, so you can probably just increase the limit.

kavigupta:
> It may be that there are so many checkboxes on California that making another React component does legitimately increase the memory usage. Not by much though, so you can probably just increase the limit.

ah ok will do

PR #1202, opened 2025-07-05
Don't enforce test duration unless test file has changed

lukebrody:


lukebrody:
Try doing a fun CSP here to group short tests together and minimize overhead

PR #1412, opened 2025-09-20
oklch in full rgb output

kavigupta:


react/src/urban-stats-script/constants/map.ts
```diff
@@ -394,3 +390,29 @@ export const cMapRGB: USSValue = {
         selectorRendering: { kind: 'subtitleLongDescription' },
     },
 } satisfies USSValue
+
+function projectBackOKLCH(h: number, s: number, l: number): [number, number, number] {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any -- this exists
```
lukebrody, line 395: https://www.npmjs.com/package/@types/color-convert

react/package.json
```diff
@@ -74,9 +74,11 @@
     "@fontsource/jost": "^5.0.18",
     "@mapbox/geojson-extent": "^1.0.1",
     "@observablehq/plot": "^0.6.16",
+    "@types/color-convert": "^2.0.4",
```
lukebrody, line 77: should be in `devDependencies`

PR #1428, opened 2025-10-02
Group together grouped stats in table

kavigupta:


react/src/page_template/statistic-tree.ts
```diff
@@ -63,8 +64,9 @@ export const statsTree: StatsTree = rawStatsTree.map(category => (
             ...group,
             contents: group.contents.map(({ year, stats_by_source }) => ({
                 year,
-                stats: stats_by_source.map(({ name, stats: s }) => ({
+                stats: stats_by_source.map(({ name, indentedName, stats: s }) => ({
                     name,
+                    indentedName: indentedName ?? undefined,
```
lukebrody, line 69:
```suggestion
                    indentedName,
```

surprised there's not a lint rule for this

kavigupta: It's null in the model!

lukebrody (approved):
One nit, otherwise looks good!

react/src/page_template/statistic-tree.ts
lukebrody: Oops the Github collapse got me

PR #1619, opened 2025-12-03
Don’t copy error screenshots in the automation

lukebrody:


lukebrody:
!updateScreenshots

lukebrody:
I didn't work because it's still executing the version on main.

Issue #1667, opened 2025-12-17
Crash caused by census being disabled on US page


kavigupta:
To replicate, go to https://urbanstats.org/comparison.html?longnames=%5B%22Baja+California%2C+Mexico%22%2C%22California%2C+USA%22%5D&universe=world&s=GczH23rwqX34iNh on a clean localStorage browser (just the default comparison page between Baja California and California), then uncheck the box for US Census and for Area and Compactness, then go to California. This should cause a crash. 

You can also replicate this by going directly to https://urbanstats.org/article.html?longname=California%2C+USA&universe=world&s=2jZmh2wde1Kqyhw

PR #1668, opened 2025-12-19
add regression test for california when no US Census

kavigupta:
This already works on the current master, but not the deployed master. "Resolves" #1667


kavigupta:
!updateScreenshots

PR #1669, opened 2025-12-19
universe selector for mapper

kavigupta:


lukebrody (changes requested):
This is an inelegant way of doing this.
We shouldn’t need to add a URL parameter.

PR #1717, opened 2026-01-14
use 3 sig figs for unitNumber

### LINKED ISSUE #1528: Number format should round when numbers are larger
In general, we do not need 3 places after the decimal in all cases, this shows up now with life expectancy by default and other statistics when you change the unit on the mapper

kavigupta:
Resolves #1528


kavigupta:
!updateScreenshots

PR #1796, opened 2026-03-05
fix preamble checkbox dissapearing when you backspace the whole thing

### LINKED ISSUE #1788: Backspacing preamble all the way in mapper makes it go away

kavigupta:
Resolve #1788

### DIFF
--- a/react/src/mapper/settings/PreambleEditor.tsx
+++ b/react/src/mapper/settings/PreambleEditor.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode, useEffect, useState } from 'react'
+import React, { ReactNode } from 'react'
 
 import { CheckboxSettingCustom } from '../../components/sidebar'
 import { locationOf } from '../../urban-stats-script/ast'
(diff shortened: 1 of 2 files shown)


react/src/mapper/settings/map-uss.ts
```diff
@@ -27,29 +32,44 @@ const pMap = { type: 'opaque', name: 'pMap', allowCustomExpression: false } sati
...
+        return undefined
+    }
+    const expr = stmt.value
+    if (expr.type === 'customNode') {
+        return expr
+    }
+    if (expr.type === 'autoUXNode') {
+        if (expr.expr.type === 'customNode') {
+            // really not sure why typescript can't figure this out, but it can't
+            return expr as PreambleAutoUXNode
```
lukebrody:
```suggestion
            return { ...expr, expr: expr.expr }
```
Type narrowing on affects the exact expression that is narrowed :sob:

react/src/mapper/settings/PreambleEditor.tsx
```diff
@@ -15,34 +24,40 @@ export function PreambleEditor({
...
-    preamble: UrbanStatsASTExpression & { type: 'customNode' }
-    setPreamble: (conditionExpr: UrbanStatsASTExpression & { type: 'customNode' }) => void
+    preamble: PreambleNode
+    setPreamble: (value: PreambleNode) => void
     typeEnvironment: TypeEnvironment
     errors: EditorError[]
     blockIdent: string
 }): ReactNode {
-    const [showPreamble, setShowPreamble] = useState(preamble.originalCode.trim() !== '')
+    const [showPreamble, setShowPreamble] = useState(shouldShowPreamble(preamble))
```
lukebrody: Do we need `useState`? Why not just store the state in the USS rather than keeping it in two places?

Issue #1482, opened 2025-10-14
add metadata to article / comparison view


kavigupta:
Add metadata to article and (maybe) comparison view.

we could just basically treat them as statistics that don't have ordinals / percentiles, giving them checkboxes on the left in a new 'metadata' section and then displaying them. Alternatively, we could place them somewhere else.

The data already exists in the .gz models, so this is a JS-only change.

PR #1837, opened 2026-03-18
add metadata columns to the article and comparison views

kavigupta:
Resolves https://github.com/kavigupta/urbanstats/issues/1482

- [x] clean up the code, this is a mess
- [x] move metadata up, it should be above the deprecated section
- [x] actual testing
  - [x] tests for metadata display
  - [x] tests for metadata sorting
  - [x] tests for metadata statname clicking
  - [x] tests for comparison where only some metadata values exist sometimes
- [x] rename the category
- [x] make the stat headers clickable, we probably don't need a statistics page, but we should link to a definition of the term
- [x] do not include statistics when none appears, the way we do with nan values in statvals


kavigupta:
!updateScreenshots

PR #1864, opened 2026-03-22
clusterMap in USS

### LINKED ISSUE #1581: Map type that clusters together points, similar to SYAU
Ideally, we should reuse the code from syau in order to do this, it should be pretty much the same thing, except that of course it will be completely customizable

kavigupta:
Resolves #1581 

- [x] weighting should be population by default
- [x] make sure stuff like editing instes works
- [x] upstream irrelevant changes
- [x] ~~allow setting the text color and possibly turning off the text (object?)~~
- [x] clean up diff
- [x] test for basic cluster map
- [x] test for weighting by something other than population
- [x] tests for alternate ramps
- [x] test for inset mode
- [x] test for opacity
- [x] ~~test for text~~


kavigupta:
!updateScreenshots

PR #1932, opened 2026-04-07
Tell Claude to use suggestions in review

lukebrody:


.github/workflows/claude-code-review.yml
```diff
@@ -60,13 +60,19 @@ jobs:
...
+                {"path": "file.py", "line": 10, "start_line": 9, "body": "comment text"}
               ]
             }
             ```
             Only include content in `body` if there is an overarching issue with the PR. Do not summarize other comments. Do not provide praise.
             If there are no comments, `body` should be "LGTM" to confirm that the review has run.
 
+            When suggesting a concrete code change, use a GitHub suggestion block in the comment body instead of a plain code block. Suggestion blocks use the exact line(s) being replaced, so the reviewer can apply them with one click. The format is:
+            ````suggestion
+            replacement line(s) here
```
lukebrody: test

PR #1906, opened 2026-04-02
add congressional representative metadata

kavigupta:
- [x] clean up diff
- [x] link to party in party symbol
- [x] party symbol should have color
- [x] congressional representatives -> Representatives
- [x] Representative number should include the end year
- [x] put the representative into election not geoid
- [x] representatives for all entites smaller than states
- [x] associate representatives with years
- [x] collapse representatives into one
- [x] [this page](http://localhost:8000/comparison.html?longnames=%5B%22MA-01+%281793%29%2C+USA%22%2C%22MA-01+%281795%29%2C+USA%22%2C%22MA-01+%281873%29%2C+USA%22%2C%22MA-01+%281877%29%2C+USA%22%2C%22MA-01+%281803%29%2C+USA%22%5D&s=FTPPbaGi1po99KPMo) crashes when you click year checkboxes it seems
- [x] scrollable view
- [x] data credit
- [x] check out why the entrypoint is increasing in size
- [x] tests for examples from various parties
- [x] tests for special elections
- [x] tests for historical parties
- [x] tests for vacancies
- [x] historical congressiona
- [x] tests for the external links
- [x] ~~test for 99-00~~
- [x] test for collapsing representatives
- [x] test for rounding
- [x] test for [this page](http://localhost:8000/article.html?longname=02139%2C+USA&s=FTPPbRwfFaNL4kJYm)
- [x] ~~test article collapse~~
- [x] ~~test comparion collapse~~
- [x] test transposed comparison collapse
- [x] test for hartford representative being stable over the 2007-2026 period
- [x] fix size.txt
- [x] remove suffixspam
- [x] thicker lines between regions
- [x] alphabetize on district name
- [x] test for reps that change while the district doesn't
- [x] test for clicking on names of districts
- [x] fix line at bottom of page
- [x] test for one of the stupid pages
- [x] when taking a screenshot, remove the scrollbar
- [x] tests for screenshots of representative tables after scrolling
- [x] tests for screenshots of representative tables after resizing
- [x] tests for keep the vertical size of the table from before to after a screenshot
- [x] test for dark mode

PR #1951, opened 2026-04-09
allow for asynchronous row loading

kavigupta:
does nothing now, but helps for future indirection being added in in #1906


kavigupta:
eh whatever let's just load on page load

PR #1988, opened 2026-04-20
congressional metadata loading (pre-table)

kavigupta:
Change that just sets up all the loading from protobufs, without the table changes


kavigupta:
!updateScreenshots

PR #1991, opened 2026-04-21
Implement search status reporting

### LINKED ISSUE #1979: Search should have some kind of loading indicator
Search sometimes takes a really long time, I assume because it's downloading or building the index. The user should get an indicator that progress is happening. Might also be nice for this to say what exactly is happening at least in a way that's legible to us (loading spinner color?) so we can qa in prod

lukebrody:
Closes #1979 

https://github.com/user-attachments/assets/6b4804d5-eab7-4fba-9795-fa14971b7052


lukebrody:
!updateScreenshots

PR #1994, opened 2026-04-21
Attempt to fix one-pixel screenshot issues in chrome

lukebrody:


lukebrody:
Doesn't work for the actual problem

PR #1996, opened 2026-04-22
basemap -> basemap options in the autoux

kavigupta:
clearer so that people don't think it being unchecked means no basemap


kavigupta:
!updateScreenshots

PR #2021, opened 2026-05-31
Implement inverse quantiles

### LINKED ISSUE #2014: Access percentiles in map script
It would be great if the custom scripts for the map could also get the percentile of a stat instead of just the stat number. Currently the only way to do this is manually assign values or manually calculate percentile, both of which crash the website.

lukebrody:
Closes #2014

Example
```
cMap(
    data=inverseQuantile(density_pw_1km, density_pw_1km),
    scale=linearScale(min=0, max=1),
    ramp=rampUridis,
    unit=unitPercentage,
    label="Population-Weighted Density (r=1km) Percentile"
)
```
benchmarks (USA, ZIP)
broadcasting (unoptimized): 44s
broadcasting (optimized): 3s ⭐ 
vector (unoptimized): 13s
vector (optimized): 3s


lukebrody:
!updateScreenshots

PR #2050, opened 2026-07-01
feat: port screenshot comparison to TypeScript, retry only failed tests

lukebrody:
## Summary
- Add `fast-png` dependency for pure-JS PNG decoding/encoding
- Port `tests/check_images.py` to `react/test/scripts/check-images.ts`, eliminating the Python dependency for screenshot comparison
- Add a JSON reporter to testcafe runs to identify exactly which individual test cases failed
- Add `failedTestNames` to the failure result type; on retry, only the failed test cases are re-run instead of the entire file

## Test plan
- [ ] Run e2e tests with `--compare` flag and confirm screenshot comparison works
- [ ] Confirm that on failure, only the failing test cases are retried (check logs for filtered case names)
- [ ] Confirm `tests/check_images.py` is no longer invoked


lukebrody:
Superseded by #2051 and #2052 which split this into separate concerns.

PR #2056, opened 2026-07-10
Temperature plots

kavigupta:
- [x] max / min should be stacked rather than just both shown, same for rain/snow
- [x] ↑ Mean high temp by month should just Mean Temp by Month when there's multiple series
- [x] Should have titles on graphs when only one geography is being plotted
- [x] horizontal gridlines
- [x] Y axis should be Rain (cm) when it's rain and Snow (rain equivalent cm) when it's snow, instead of percipitation
- [x] link the histogram toggles
- [x] clip to the x axis
- [x] bar charts
- [x] tests for showing just high temperature and just low temperature on individual articles
- [x] tests for showing both temperatures on individual articles
- [x] tests for showing both temperatures on a comparison
- [x] tests for showing just high temperature on an individual article
- [x] tests for high and low temperature histograms
- [x] tests for comparisons between high and low temperature histograms (e.g., /comparison.html?longnames=%5B"San+Luis+city%2C+Arizona%2C+USA"%2C"Utqiaġvik+city%2C+Alaska%2C+USA"%5D)
- [x] tests for just rain and just snow
- [x] tests for snow when there is none
- [x] tests for snow comparsion when there is none on one
- [x] tests for both rain and snow in an individual article
- [x] tests for both rain and snow in comparsion
- [x] desloppify everything

**Before:**

```
125     consolidated
178     quiz_sampling_info
685     shape
788     order
1388    data
3293    total
```

**After:**

```
125     consolidated
178     quiz_sampling_info
685     shape
788     order
1412    data                   [+2%]
3289    total                  [0%]
```


kavigupta:
!updateScreenshots

kavigupta:
!updateScreenshots

kavigupta:
!updateScreenshots

PR #2069, opened 2026-07-11
Let named-arg documentation use HumanReadableName

lukebrody:
## Summary
- `Documentation.namedArgs` now maps argument name to `HumanReadableName` (`string | HumanReadableElement[]`) instead of plain `string`, so per-argument docs can use subscript/superscript/code-span elements like everywhere else.
- Updates the two places that render these descriptions: the "Named Arguments" list in `uss-documentation.tsx`, and the checkbox label in `AutoUXEditor.tsx`, to go through `reifyReact`.
- Widens `CheckboxSettingCustom`'s `name` prop from `string` to `ReactNode` so it can accept the reified content.
- Existing string-valued `namedArgs` entries throughout `constants/*.ts` are unaffected — `string` is still a valid `HumanReadableName`.

## Test plan
- [x] `npx tsc --noEmit`
- [x] `npx eslint` on changed files
- [x] `npm run test:unit -- --test='unit/*.test.ts'` (341 tests, only pre-existing network-dependent failure in `random.test.ts` unrelated to this change)


lukebrody:
This doesn't make any sense

PR #2074, opened 2026-07-12
Give plot x-axis labels more room below the axis

kavigupta:
## Summary
- `PlotComponent`'s bottom margin was too tight, causing the x-axis title to sit too close to the tick labels. Bumps `marginBottom` from `40`/`80` (non-transposed/transposed) to `55`/`90`.
- Applies to every plot rendered through `PlotComponent`, including the existing population-density `Histogram`.

## Test plan
- [x] `npx tsc --noEmit` clean
- [x] `npx eslint . --ignore-pattern=src/data --max-warnings=0` clean
- [ ] Visually confirm the x-axis label has better spacing on the density histogram, in both normal and transposed/vertical layouts


kavigupta:
!updateScreenshots

PR #2090, opened 2026-07-18
Add kelvin temperature unit

lukebrody:


lukebrody:
Testing that the automated reviewer would catch this sort of issue

PR #2101, opened 2026-07-21
fix + vs ! alignment when both appear

### LINKED ISSUE #2086: + and ! buttons don't line up
Image

kavigupta:
Resolves #2086


kavigupta:
!updateScreenshots

PR #2102, opened 2026-07-21
fix giant footer in histogram plots

### LINKED ISSUE #2081: Histograms have a very large footer
Image

This is problematic

kavigupta:
Resolves #2081


kavigupta:
!updateScreenshots

PR #2119, opened 2026-07-26
Editable percentiles on articles/comparisons

### LINKED ISSUE #2114: Editable percentiles on articles/comparisons
Like we should be able to click the percentile and change it to be 50 and it'll jump to a geography that's closest to 50 percentile

We probably want something like it never picking the current one if possible, so you don't have a frustrated user wondering why it's not working when trying to jump to the 99th percentile by population from California for example

kavigupta:
Closes #2114.

Makes the population **percentile** column on article and comparison tables editable, mirroring the existing editable ordinal. Entering a percentile navigates to the **bottom of that bucket**: the lowest-value geography that still has at least that share of the population below it. So:

- **0th** → least-valued geography (last)
- **100th** → most-valued geography (first)
- **50th** → population-weighted median

### Notes
- The lookup is extracted into a pure, unit-tested `percentileBucketIndex` (`react/unit/percentile-navigation.test.ts`).
- Population percentiles cap below 100 (the most populous geography holds a large share of the population), so e.g. the 95th percentile among US states resolves to California.
- `alwaysSubmitOnEnter` was added to the editable field so submitting the *currently-displayed* percentile still navigates — e.g. the 0th percentile from North Dakota, which shares the 0th bucket but isn't the least-populous state — while blur stays guarded against accidental navigation.
- The field resets (via remount) when an edit resolves to staying put.

### Stacked
Based on #2121 (relocating the Percentile component into table.tsx), itself on top of the merged #2118 (percentileSuffix extraction). Review/merge #2121 first; GitHub will retarget this to `main` once it lands.

### Testcafe screenshots
The percentile now renders as an editable box, so reference screenshots will change — regenerate with a `!updateScreenshots` comment.


kavigupta:
!updateScreenshots

kavigupta:
!updateScreenshots

PR #2120, opened 2026-07-26
Move the checkbox controls out of the sidebar

lukebrody:
`sidebar.tsx` had quietly become the module everything imports checkboxes from. Nine of its consumers — the mapper's four settings editors, the quiz result, so-you're-an-urbanist, the histogram, the related buttons, and the statistic tree — import `CheckboxSetting` or `CheckboxSettingCustom` and nothing else from it, and none of them render anywhere near a sidebar. `template.tsx` is the only consumer that actually wants `Sidebar`.

So the checkbox controls move to their own `components/checkbox-setting.tsx`, and `sidebar.tsx` goes back to being about the sidebar.

This is a pure move — no behavior change. The only difference to the components themselves is that `CheckboxSettingJustBox` and its props interface are now exported, since they're part of what a checkbox-primitives module offers rather than a sidebar implementation detail.


lukebrody:
Added a second commit: names the two staging idioms this code was spelling out at each use site — whether staging is changing a setting, and what value it currently has.

That lets the stat tree's group checkbox read its value and its highlight from one subscription instead of two, and lets the temperature control share the highlight styling rather than rebuilding it. `useStagedSettingKeys` becomes private behind `useIsStaged`, since nothing needed the keys themselves.

The file move in the first commit is unchanged.

PR #2121, opened 2026-07-27
Restructure Percentile: relocate to table.tsx and render from composed pieces

kavigupta:
Two behavior-preserving restructurings of the `Percentile` component, split out ahead of the editable-percentiles feature (#2114) so that PR contains only the editable behavior:

1. **Relocate** the (non-editable) `Percentile` component from `display-stats.tsx` into `table.tsx`, next to the analogous `Ordinal`.
2. **Render from composed pieces** — instead of dropping `percentileText`'s whole string into the div, render the number and the "`<suffix> percentile`" / "`%`" as separate elements. Output is identical; this just makes the number a standalone element so it can later be swapped for an editable field. `percentileText` is still used to size the column.

No rendering changes.


kavigupta:
!updateScreenshots

PR #2128, opened 2026-07-29
Yesterday option for the friends panel toggle

### LINKED ISSUE #2122: Yesterday should be an option for the toggle for the friends panel on juxta

kavigupta:
Closes #2122. Rebased on top of #2129, which extracted `ViewMode` / `viewModeOptions` and fixed retrostat's current-period label.

Adds a third option to the friends panel toggle on the quiz results page, showing everyone's scores on the preceding quiz.

- Labels follow the quiz kind: **Yesterday** for juxtastat, **Last Week** for retrostat.
- Friends' scores come from the existing `/juxtastat/todays_score_for` endpoint with the day/week identifier decremented, so there's no server change.
- Our own previous result is read out of the local quiz history (`ourPreviousResultToDisplayForFriends`). A quiz we didn't finish reports as "Not Done Yet", matching how the server reports friends' unfinished quizzes.

Uses `historyKeyForTimeIdentifier` from #2130.

Two things this deliberately does *not* do:

- Clicking the toggle refetches the summary stats along with the scores, even though they don't depend on the period. That's one redundant POST on a user-initiated action, which beats splitting the effect in two.
- The effect can still race on itself: click the toggle twice inside one request window and a late response can render the wrong period's scores until the next fetch. That race already exists on `main` (the effect re-runs when the friends list changes) — the toggle just makes it easier to reach, so it belongs in its own fix rather than here.

## Testing

`<kind>-friends-previous-quiz` and `<kind>-friends-previous-quiz-not-done` added to the friends test template, so they run for juxta and retro, desktop and mobile. The first has Alice and Bob play the previous quiz *before* today's, then checks that the toggle shows the previous quiz's patterns and switches back.

Reference screenshots will need `!updateScreenshots`: every friends panel screencap gains a third toggle button, and the new test adds one screenshot.

## Known limitation (pre-existing)

`reportToServerGeneric` only uploads history entries newer than `MAX(day)` already stored for the user. Someone who plays out of order (today's quiz before yesterday's) never uploads the older one, so they'll show as "Not Done Yet" in the Yesterday view. Playing in order — the normal case — is unaffected. Fixing it means changing the sync protocol, so I left it alone.


kavigupta:
!updateScreenshots

kavigupta:
!updateScreenshots

PR #2129, opened 2026-07-29
Extract the friends panel view mode options

kavigupta:
Small refactor of the friends panel toggle, split out of #2128 so that PR is mostly the feature itself.

- Names the toggle's state `ViewMode` instead of repeating the string union inline.
- Moves the toggle options into `viewModeOptions(quizKind)`, so the labels live in one place and are chosen per quiz kind.

One visible change: retrostat's current-period button now reads **This Week** rather than "Today", which was wrong for a weekly quiz. Nothing else changes, and no test asserts on that label (the one test that clicks a button with the exact text "Today" is a juxtastat test).

Retro friends panel reference screenshots will need `!updateScreenshots` for the label.


kavigupta:
!updateScreenshots

PR #2132, opened 2026-07-30
Pin plot tooltips by clicking them

### LINKED ISSUE #2082: Tooltip should show up in the screenshot for a histogram
They should also be a bit more persistent, being part of the state even if you click off somewhere else. have a little "x" box to dismiss them. Not in anything super persistent, it should dissapear when we navigate to another page / reload

kavigupta:
Resolves https://github.com/kavigupta/urbanstats/issues/2082


kavigupta:
!updateScreenshots

PR #2133, opened 2026-07-31
Place tooltips against the frame, and clear of the legend

### LINKED ISSUE #2083: Tooltips should be below the line when it's above the halfway point on the graph
Otherwise it'll intersect with the frame

Image

kavigupta:
Fixes #2083.

Plot picks the corner a tooltip hangs off by fitting the box inside the **SVG** — its `fitBottom` test is `y - h - m - r*2 > 0`, measured against the SVG's top rather than the frame's. Our plots have `marginTop: 80`, so for a point at the top of the frame Plot sees 80px of "room" above, takes its default `preferredAnchor: "bottom"`, and draws the tooltip up into the margin: across the frame and under the settings bar.

## What this does

New `plots-tip-anchor.ts` picks the anchor from the **frame** instead, via a render transform on the tip mark:

- the box grows towards the middle — below a point in the top half, above one in the bottom half — and is pushed inwards at the sides when a centred box would overhang;
- the legend is drawn inside the frame in the top left, which is exactly where the left end of a cumulative histogram lives, so a tooltip that would land on it is **dropped below it** (`mark.dy`) and joined back to its point by a leader with an arrowhead;
- the leader passes **under** the legend, which is opaque chrome: where the point it runs up to is itself behind the legend, only the stub between the legend's foot and the tooltip shows;
- if the displaced tooltip would not then fit above the frame's foot, it stays where it was rather than being pushed off the plot.

## Notes for review

- **It sets `anchor`/`dy` on the mark from inside the render transform.** `Tip.render` reads `const {anchor} = this` at render time, which is what makes a per-point anchor possible at all — the anchor is otherwise per-mark, while the pointer-driven tip draws whichever point you are on. That is an internal of Plot's tip mark; if it goes away, tooltips fall back to Plot's own fitting rather than breaking. `preferredAnchor` is the public alternative but cannot express the rule, and would not fix this case anyway: near the left edge `fitRight` is false, so Plot takes the `left` branch and centres the box vertically on the point, still crossing the frame.
- **The box is measured up front on a canvas**, because Plot only measures a tooltip once it is in the document — after the anchor has to be chosen. Same trick `manualLegend` already uses on its own labels. The em→px assumption (`plotFontSize = 16`) only feeds placement decisions, so being off means a slightly different anchor, not a broken layout.
- **The leader is appended a frame later**: anything added to the tip's group before Plot's own layout pass gets mistaken for part of the tooltip.
- `manualLegend` now returns `{ marks, bounds }` and `SeriesPlot` builds it *before* `buildPlot`, so the bounds can be threaded to the tips — hence the extra parameter on `buildPlot`, `transposeAwareTip` and `seriesTip`.

## Testing

`histogram-tooltip-top-of-frame` hovers the left end of a cumulative relative histogram on a four-country comparison — 100%, so the top of the frame *and* its left edge, with a five-line tooltip — and asserts the leader exists, that the tooltip clears the frame's highest gridline, and that its rect does not intersect the legend's. It expands the GHS-POP row rather than taking the first expand toggle, which is a census row only the US and Canada have data for (and a lone series draws no legend to collide with).

The full histogram suite passes locally (43 tests). The reference strings were regenerated with `checkString` off; they differ from main by exactly one line each — the legend's `data-test-id`, which the test needs to find it by — and the one plot with no legend is unchanged.

The new screenshot still needs `!updateScreenshots` — no reference exists for `histogram-tooltip-top-of-frame` yet.


kavigupta:
!updateScreenshots

PR #2140, opened 2026-08-02
Move article warnings above table content

lukebrody:


lukebrody:
actually going to just revamp the article warning system

PR #2141, opened 2026-08-02
(wip) Select mode for article and comparison views

lukebrody:


lukebrody:
split into !2153 and friends

PR #2142, opened 2026-08-02
Move warnings for articles/comparisons into the table

lukebrody:
This is cleaner generally, and also sets us up well for edit mode


lukebrody:
!updateScreenshots

PR #2156, opened 2026-08-05
04/07: Choose statistics on the article table itself

lukebrody:
Stacked on #2155. The core of the stack.

Picking which statistics to see meant hunting through a sidebar tree that sat nowhere near the table it controlled, and on mobile through a modal that covered the very table you were trying to judge the effect on. Neither showed you what a checkbox would do until after you clicked it.

Edit mode puts the category/group tree in the table's left header column, with each statistic's real value beside its checkbox, so the selection is made against the data it produces. It is ephemeral — nothing about it is persisted.

**Decisions worth reviewing**

- It replaces the mobile modal outright, which is why `Modal.tsx` goes here rather than in the deletion PR — the modal and the edit button occupy the same cell, so they can't coexist.
- The sidebar's tree deliberately **stays** for now. The comparison table still depends on it until 05/07, so deleting it here would leave comparison pages with no way to select statistics. It comes out in 06/07.
- While both trees exist, `link_settings_test_template` compares highlighted inputs as a set rather than in document order, and expects both trees' checkboxes. 06/07 restores the order-sensitive form once only one tree remains. Same reason for the temporary `uncheckAllSidebarCategories` split in `test_utils`.

**Caveat:** typecheck, knip and lint are clean, but I have not run the e2e suites locally. Screenshots will need regenerating.


lukebrody:
!updateScreenshots

PR #2157, opened 2026-08-05
05/07: Choose statistics on the comparison table too

lukebrody:
Stacked on #2156.

The comparison table gets the same edit mode, sharing the tree and its state (`useCategoryTreeState`) so the two tables can't disagree about what a checkbox means.

Two things differ here, both forced by the comparison's layout:

- The Select button goes in the super header's left spacer rather than the top-left cell, which on a comparison is too narrow to hold both a button and the column's name. It reads "Select Statistics", since the column headers already offer editing the regions being compared.
- Editing always uses the untransposed orientation — the tree runs down the left column, and there are almost always more statistics than regions. This is the extra input to `comparisonTableShape` that 02/07 set up for.

`ComparisonEditTable` is split out so the column measurement, which runs over every statistic rather than the selected ones, only happens while edit mode is actually open.


lukebrody:
!updateScreenshots

PR #2158, opened 2026-08-05
06/07: Remove the sidebar's copy of the statistic tree

lukebrody:
Stacked on #2157. Mostly deletions.

Now that both tables carry the selection themselves, the sidebar's tree is a second set of controls for the same settings, sitting where it can't show what any of them do. Out go `StatsTree`, `Years`, `SidebarForStatisticChoice`, the CSS and z-index layers that supported them, the `pointing` prop on `ExpandButton` (the sidebar tree was its only `'left'` caller), and the tests and 23 reference screenshots that covered them.

The sidebar keeps the settings that aren't about which statistics to show.

This is also where the transitional test scaffolding from 04/07 comes out: `link_settings_test_template` goes back to comparing highlighted inputs in document order against the edit tree alone, and `uncheckAllSidebarCategories` disappears.


lukebrody:
!updateScreenshots

PR #2159, opened 2026-08-05
07/07: Let a warning take you to the setting it is about

### LINKED ISSUE #1692: edit mode for articles / comparison pages
In this mode, the tree-structured checkbox array should be replicated on the table itself, allowing users to check/uncheck specific statistic categories or groups.

Categories should appear collapsed if they are unchecked

Groups should be greyed out but otherwise appear normally if they are unchecked

lukebrody:
Stacked on #2158. Last of the stack; the tip is identical to the `article-edit` branch.

A warning already names what to do about itself — "Select 2010", "Enable one" — but left the user to go find where. This makes that phrase the button: it opens edit mode, where the years and sources it refers to are.

Two consequences:

- The edit tree's own warnings stay plain text, since they are already where the button would send you. That's what the optional `onEdit` on `warningContent` distinguishes.
- On the edit tree, warnings move from standing in for a row to sitting in the row the group already has — the tree lists every group whether or not it has anything to show, so there's no missing row to stand in for. A warning covering a whole category is repeated on each of its groups, since they're rows apart on the tree.

Also rewords the empty state from "No Statistic Categories are selected" to "No Statistics are selected".


lukebrody:
!updateScreenshots

PR #2179, opened 2026-08-09
Keep the table view's controls out of its screenshot

### LINKED ISSUE #2178: UI elements visible in screenshot of table view mode
Image

lukebrody:
Fixes #2178.

The statistic/table page's screenshot captures `headersRef`. That element had
accumulated the add-column box and the Filter / Edit Table and Convert to Map
buttons, so the exported image showed interactive UI that means nothing in a
static picture.

Rather than teaching those controls to hide themselves in screenshot mode, this
moves them out of the captured element. Nothing involved has margins, so the
rendered page is byte-identical, and there's no extra screenshot-mode state to
keep in sync as controls come and go. The header's cross-source-border
disclaimer moves out with them — it already renders as a footnote during a
screenshot, so it was never captured from there anyway.

Already covered by e2e: the `downloadImage` screenshots in
`cross_source_border_disclaimer.test.ts` and `statistics.test.ts` capture this
exact image, and their current references show the bug. **Needs
`!updateScreenshots`** once CI has run.


lukebrody:
!updateScreenshots

PR #2183, opened 2026-08-09
Pin docker base images

lukebrody:
Pin docker base images to cache correctly and match package repo snapshot for Chrome 143


lukebrody:
!updateScreenshots

lukebrody:
!updateScreenshots

Issue #2204, opened 2026-08-11
unit inference


kavigupta:
this is a particular issue for captions, e.g., the USS program

```
condition (commute_bike < 0.1)
cMap(data=density_pw_1km, scale=linearScale(), ramp=rampUridis)
```

renders the label as

```
PW Density (r=1km) where Commute Bike % < 0.1
```

which is incorrect.

unit inference fixes this and also makes it so that the unit can be inferred in situations like

```
cMap(data=density_pw_1km/density_pw_2km, scale=linearScale(), ramp=rampUridis)
```

to be unitless and not PW density.

I think the approach should be relatively straightforward, just a set of abstract interpretation semantics. the types are

- standard unit, this is represented as a sparse vector like {m: 2, s: -1} for square meters per second. unitless is represented as {}
- percent
- unknown unit, this is used for constants, etc

the semantics are

- standard `op` standard -> normal unit semantics. if uninferrable, e.g., with + or -, goes to unknown unit.
- percent `+/-` percent -> percent
- percent `otherop` percent -> {}
- percent `op` standard -> treat percent as {} and then do the normal unit semantics
- unknown unit `op` anything -> anything

Outstanding tasks

- [x] handle constants properly
- [x] make sure temperature - 5 properly doesn't typecheck
- [x] make sure backwards inference works properly and constants are correctly annotated
- [x] implement unit rules for basic built in functions
- [x] casting population / fatalities to number in certain cases
- [x] make sure temperature and temperature differences are handled correctly
- [x] double check that percentage semantics work properly
- [x] `log(x [in people/km^2])` in captions
- [x] Unit mismatch should lead to a warning
- [x] something like hospital_distance + rainfall * sun_hours shouldn't unit check

PR #2215, opened 2026-08-13
Make units abstract, and choose how to write one by searching its ladders

kavigupta:
Groundwork for #2204. Stacked on #2229; its contents drop out of this diff once that merges.

Each `UnitType` had its own `renderValue` in `unit-display.tsx` — a 400-line switch — so what a density is, and the fact that it is stored per square kilometer, existed only inside the arithmetic of one branch of it. Nothing could ask a unit anything about itself, which is what #2204 needs.

## Units are abstract

```ts
export type BaseUnit = 'person' | 'm' | 's' | 'usd' | 'fatality' | 'g'

export type Unit = (
    { kind: 'unknown' }
    | { kind: 'raw-percentage', partyColor?: Hue }
    | { kind: 'delta-percentage', partyColor?: Hue }
    | { kind: 'lead-percentage', partySystem: 'democratic' | 'left' }
    | { kind: 'temperature-F' }
    | { kind: 'dimensionfull', scales: { baseUnit: BaseUnit, power: number }[] }
)
```

A unit says what kind of quantity a number is and nothing about how it is written. How a particular column of numbers is stored is separate, since the same quantity reaches us in different units — an elevation in meters and a distance in kilometers are both lengths:

```ts
export interface StoredUnit { unit: Unit, toBaseUnits: number }

density: dimensionfull({ person: 1, m: -2 }, 1e-6),   // stored per square kilometer
minutes: dimensionfull({ s: 1 }, 60),
```

## Writing one is a search

Each base unit has a ladder of the units it can be written in, and a quantity is written by choosing one per base unit — whichever combination costs least once the number it leaves behind is counted, at one per digit past the third before the point and nine tenths of one for the point and each leading zero after it. Two costs say what a unit is worth beyond the digits it saves: an abbreviation costs one digit's worth, so a population runs to four digits before it is counted in thousands; a coarser unit costs half of one.

Which system the reader measures in picks the ladder rather than filtering one, so imperial is not a special case anywhere in the search.

Composition alone gets the units wrong: nothing in a ladder can say that people per area belong in square kilometers while an area itself belongs in square meters until it grows, since both are the same step up the same ladder — and no cost for a leading zero satisfies both at once (`0.081 km²` needs one under 0.5, `12 345 /km²` needs one over 1.5). So each dimensions may name the unit it is conventionally written in, which costs nothing:

```ts
'm^-2 person^1': { m: 'km' },
'g^1 m^-3': { g: 'μg', m: 'm' },
'fatality^1 person^-1': { person: '100k' },
```

## What changes

A quantity is written in whichever unit reads best rather than the one it happened to be stored in, so:

| | before | now |
|---|---|---|
| elevation 1234 m | `1 234 m` | `1.23 km` |
| mean distance 0.5 km | `0.50 km` | `500 m` |
| commute time 34 min | `34` | `34.0 min` |
| a duration of 1234 hours | `1234:00` | `51.4 days` |
| area 0.5 km², imperial | `124 acres` | `0.193 mi²` |
| $1 234 | `$1.23k` | `$1 234` |

Lengths and durations follow from `{m: 1}` and `{s: 1}` each being one unit: an elevation and a distance are the same kind of quantity, so they are written the same way. Acres are the one thing composition cannot express, an acre being no power of any length. **Reference screenshots need regenerating for these.**

Everything else is unchanged, including all the percentages, and values that do not occur (pollution above 10 000 μg/m³, densities above a million per km², fatality rates above 0.1 per person).

## Rendering

`renderQuantity(value, storedUnit, colors, settings)` is the whole of the React side. A number is written as the pieces the browser shapes separately — a party, a plus and a size; a sign and a number; the hours and minutes of a duration — because joining them into one string moves a pixel, as #2229 found.

## Test plan

- All 30 unit types × 16 values × both measurement systems × both temperature units, rendered to markup with `renderToString` and diffed against main. The differences are the table above, plus 30 lines where rainfall's unit is one run of text rather than two.
- `unit/unit-display.test.ts` covers each unit type at several magnitudes in both systems, zero, negatives, NaN, the abbreviation boundaries, celsius, the party colors, and the margin inequalities.


kavigupta:
Superseded. Everything here landed in pieces, and several parts differently: units come from one pool rather than per-base-unit ladders, so an acre is expressible where this dropped it; conventions belong to the statistic rather than to a dimension key, so a derived quantity floats; `Unit` is a temperature or a scalar, with `difference` in place of seven kinds; the search is its own module taking its pool as an argument; and a duration stays a quantity of time.

#2250, #2251, #2252, #2255, #2256, #2257, #2265, #2267, #2270, #2272, #2277, #2278, #2279, #2281, #2282, #2292, #2295, #2296.

PR #2216, opened 2026-08-13
Infer the unit of a computed quantity

### LINKED ISSUE #2204: unit inference

kavigupta:
Closes #2204. **Stacked on #2215**, which introduces the unit framework this builds on and changes no behavior; review that one first, and this diff shrinks to the feature once it merges.

A map or a caption could only say what unit its data was in when the data was a statistic column. Anything computed fell back to `classifyStatistic` on the words of the label, so `density_pw_1km / density_pw_2km` was labelled a density, and the condition in

```
condition (commute_bike < 0.1)
cMap(data=density_pw_1km, scale=linearScale(), ramp=rampUridis)
```

was written as `PW Density (r=1km) where Commute Bike % < 0.1`.

## Inference

`urban-stats-script/unit-inference.ts` infers the unit of an expression by abstract interpretation over the AST, over the `Unit` from #2215 (dimensions + the multiplier into base units), with `undefined` for what cannot be inferred:

- dimensions and multipliers follow the arithmetic, so `population / area` is exactly the unit a density is stored in, and `area ** 0.5` is kilometers
- adding quantities of different dimensions, or of the same dimensions stored differently (`hospital_mean_dist + elevation`), gives no unit
- a presentation survives an operation that keeps the quantity the same kind of thing: percentage ± percentage is a percentage, percentage × percentage is a plain fraction, percentage × population is a number of people
- an unknown operand leaves the other one's unit alone, so `population * 2` is still a population
- comparisons and logical operators produce booleans, which have no unit

`deriveMapUnit` and `deriveTableColumnUnit` read the unit off the map's `data` or a column's `values` the same way the label is derived, and are used when nothing was passed explicitly. Where inference gives nothing, the old label-text heuristic still applies.

## Captions

A constant is written in the unit of what it is compared against, in whichever display unit reads best:

| | before | after |
|---|---|---|
| `commute_bike < 0.1` | `Commute Bike % < 0.1` | `Commute Bike % < 10%` |
| `ped_cyclist_fatalities_per_capita > 1e-5` | `... > 1x10^-5` | `... > 1/100k` |
| `commute_time_median > 600` | `... > 600` | `... > 10:00` |
| `population > 1000000` | `Population > 1m` | unchanged |
| `elevation > 1234` | `PW Mean Elevation > 1 234` | unchanged |

The unit is only named when the number has to be rescaled to be read in it, so quantities already stored in the unit they are displayed in keep their bare number.

## Durations past a day

The one display change: `time^1` gains days and years above a day, so a duration that outgrows h:mm is readable. No statistic reaches it — commute times are minutes and sunny hours are under 24 — but a computed one can.

## Test plan

- `unit/urban-stats-script-unit-inference.test.ts`: 50 cases over the inference rules, the unit arithmetic, and `deriveMapUnit` / `deriveTableColumnUnit` rendered end to end.
- `unit/urban-stats-script-derive-human-readable-name.test.ts`: the caption cases above, including the ones that must not change.
- Screenshots should only move for a mapper colorbar whose data is a computed expression: every condition in the e2e mapper scripts is on `population`, and the one on `white` belongs to a map that passes its own label.

### DIFF
--- a/react/src/components/display-stats.tsx
+++ b/react/src/components/display-stats.tsx
@@ -2,16 +2,15 @@ import React, { CSSProperties, ReactNode } from 'react'
 
 import { useColors } from '../page_template/colors'
 import { useSetting } from '../page_template/settings'
-import { classifyStatistic, UnitType } from '../utils/unit'
+import { classifyStatistic, Unit, unitTypeToUnit } from '../utils/unit'
 
-import { getUnitDisplay } from './unit-display'
+import { getQuantityDisplay } from './unit-display'
 
-export function Statistic(props: { style?: React.CSSProperties, statname: string, value: number, isUnit: boolean, unit?: UnitType }): ReactNode {
+export function Statistic(props: { style?: React.CSSProperties, statname: string, value: number, isUnit: boolean, unit?: Unit }): ReactNode {
     const [useImperial] = useSetting('use_imperial')
     const [temperatureUnit] = useSetting('temperature_unit')
 
-    const statisticType = props.unit ?? classifyStatistic(props.statname)
-    const unitDisplay = getUnitDisplay(statisticType)
+    const unitDisplay = getQuantityDisplay(props.unit ?? unitTypeToUnit(classifyStatistic(props.statname)))
     const { value, unit } = unitDisplay.renderValue(props.value, useImperial, temperatureUnit)
 
     return (
--- a/react/src/components/load-article.ts
+++ b/react/src/components/load-article.ts
@@ -10,7 +10,7 @@ import { findAmbiguousSourcesAll, statParents, StatName, StatPath, statPathToOrd
 import { assert } from '../utils/defensive'
 import { HumanReadableName } from '../utils/human-readable-name'
 import { Article, CongressionalRepresentativeTable, ICongressionalRepresentative, ICongressionalRepresentativePointer, IFirstOrLast, IMetadata } from '../utils/protos'
-import { UnitType } from '../utils/unit'
+import { Unit } from '../utils/unit'
 
 import { CountsByUT, forType } from './countsByArticleType'
 import { electionDisclaimerForRow, type Disclaimer } from './disclaimer-text'
@@ -111,7 +111,7 @@ const dataCreditExplanationPageByMetadataIndex = new Map<number, string>(
 interface StatisticCellRenderingInfoCommon {
     articleType: string
     statname: HumanReadableName
-    unit?: UnitType
+    unit?: Unit
     statpath?: StatPath
 }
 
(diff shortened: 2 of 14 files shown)


kavigupta:
Closing: this has been landed as a chain of small PRs off main rather than as one branch. What replaced it, roughly in order:

**The quantity itself** — #2249, #2250, #2251, #2252, #2256, #2281, #2282, #2292, #2295, #2296: a unit becomes dimensions plus a decoration, a statistic carries the units it is written in, and everything that writes a number goes through one function.

**Writing one** — #2233, #2236, #2279, #2286, #2324, #2345, #2349, #2366: N/A whatever the unit, the space after a bare solidus, rainfall in the reader's small length, embed-card spacing, per-capita statistics, legend caption spacing, and the default arguments that hid which reader a caller meant.

**Inferring one** — #2300, #2303, #2304, #2305, #2311, #2320, #2340: the algebra, the forward walk, the search's cost model, maps and tables, what each built-in makes of a quantity, a caption's numbers in the units they are read from, and `maximum(x, 80)`.

**Declaring one** — #2350: statistics declare their units in Python instead of the frontend guessing from the display name.

#2348 has what is left: a caption's numbers written in the units of whoever is reading the page.

PR #2220, opened 2026-08-14
Add keyboard navigation to the universe dropdown

lukebrody:
Picking a universe no longer requires the mouse. Opening the dropdown focuses the search box, arrow keys move a highlight through the filtered options, and Enter selects the highlighted one.

The mouse and the keyboard share a single `highlighted` index rather than each having their own notion of the active row. That is why the option background is set inline instead of via `:hover` — with a CSS hover rule you would see two highlighted rows at once whenever the pointer sat over a row the keyboard had moved off of.

Enter is ignored when focus is already inside the options list, since the browser will fire the click itself.

Stacked on #2219.


lukebrody:
!updateScreenshots

PR #2221, opened 2026-08-14
Put the universe selector in the header's right panel on mobile

lukebrody:
The universe selector now sits next to the screenshot and CSV buttons at every screen size. Mobile previously swapped it into the left panel in place of the logo, so it had a different position and different neighbours depending on the viewport, which made it hard to style once.

The consequence is that mobile shows no logo while a universe is active — the left panel is just the hamburger. That is intentional; the selector is the more useful thing to have in that space.

Stacked on #2220.

### DIFF
--- a/react/src/common.css
+++ b/react/src/common.css
@@ -71,6 +71,11 @@ button, select {
     cursor: pointer;
 }
 
+button.borderless {
+    border: none;
+    background-color: transparent;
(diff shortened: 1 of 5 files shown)


lukebrody:
This produces an invalid layout, so we'll just merge it with #2223

PR #2222, opened 2026-08-14
Keep the universe dropdown inside the viewport

lukebrody:
The dropdown is `width: 500%` of its anchor with no left offset, which is 432px hanging off an 86px flag button. On a phone that runs off the edge of the screen and widens the whole page.

CSS alone cannot fix this, since the fit depends on where the anchor lands, so the placement is measured at open time: cap the width at the viewport and shift the dropdown left far enough to fit, leaving a 5px margin on both sides. It re-measures on resize while open.

#2221 is closed and its commit is reverted on this branch, so the diff against main is just the placement fix. That header layout change now lives in #2223.


lukebrody:
!updateScreenshots

PR #2223, opened 2026-08-14
Restyle the universe selector as a labelled button

### LINKED ISSUE #2203: Restyle universe screenshot, and CSV buttons match other interactive elements
Image

These would act more like normal buttons.

Plus, universe would be a labeled concept.

lukebrody:
Closes #2203.

The universe selector is now a real button that fills the header bar's height, with a "UNIVERSE" label under the flag. A bare flag did not read as clickable, and nothing on the page explained what a flag in the header meant.

It is rounded at the bottom corners only, so it reads as hanging off the top of the page rather than floating in it. The screenshot button gets corners matching its icon for the same reason, and `.borderless` picks up a 7px radius and a transparent background so the header buttons agree with each other.

This also carries the header placement change from #2221: the selector sits next to the screenshot and CSV buttons at every screen size, rather than swapping into the left panel in place of the logo on mobile. That change produced an invalid layout on its own, which is why #2221 is closed and reverted out of #2222 and lands here instead. The consequence is that mobile shows no logo while a universe is active — the left panel is just the hamburger. That is intentional; the selector is the more useful thing to have in that space.

This is also where the screenshot and CSV buttons lose their inline `background: none` and start lighting up on hover like every other button. That inline value outranked the shared `button:hover` rule, which is why they were flat before.

`.hgap` drops from 1em to 0.75em. That is a site-wide spacing change rather than a header-only one, folded in here because it came out of the same visual pass.

Last of the chain, stacked on #2222.


lukebrody:
!updateScreenshots

PR #2233, opened 2026-08-14
Say N/A for a quantity we do not have, whatever it is measured in

kavigupta:
Third of six, splitting #2230. Stacked on #2232.

A missing value read three different ways, depending on which unit it would have been in:

| | before | after |
|---|---|---|
| most units | `NaN` | `N/A` |
| money | `$NaN` | `N/A` |
| a count of hours | `NaN:NaN` | `N/A` |
| written in a party's color | `N/A` | unchanged |

The check now lives in one place — the helper every unit's number goes through — rather than being left to whatever `toFixed` happened to produce.

## What changes on screen

48 of the 1440 cases in the sweep, all of them a missing value. Which screenshots move depends on which statistics are missing for the geographies they show. **This needs `!updateScreenshots`.**

## Test plan

- The sweep against `main` shows the 48 missing-value differences and nothing beyond what #2231 and #2232 already account for.
- `unit/unit-display.test.ts` checks nine unit types render a missing value as `N/A`.


kavigupta:
!updateScreenshots

PR #2236, opened 2026-08-15
Set a solidus with nothing in front of it with a space after it

kavigupta:
Last of six, splitting #2230. Stacked on #2235.

A density read `/ km²`, with a non-breaking space, and a rate per a hundred thousand people read `/100k`, without one. The space belongs to the kind of solidus rather than to the unit: one between a numerator and a denominator is set tight, and one standing for a bare "per" is set with a space.

| | before | after |
|---|---|---|
| density | `1 234 / km²` | unchanged |
| per capita | `1.20/100k` | `1.20 / 100k` |
| pollution | `8.20 μg/m³` | unchanged |
| rainfall | `120.0 cm/yr` | unchanged |

## What changes on screen

The fatality-rate rows, which are the only bare-denominator unit that was set tight. **Needs `!updateScreenshots` for the article rows showing traffic and pedestrian fatality rates.**

## Test plan

- The sweep against `main` accounts for every difference; the per-capita rows now differ by the space as well as by #2232's grouping.
- `unit/unit-display.test.ts` pins both kinds of solidus.


kavigupta:
!updateScreenshots

PR #2241, opened 2026-08-15
Write each number as one string rather than several

kavigupta:
New first PR in the #2231–#2236 stack. **Only rasterization changes; not one character does.**

A margin was written as its party, a plus and its size, in three children:

```tsx
<span style={spanStyle}>
    {emphasis.labels[side]}
    +
    {magnitude.toFixed(leadPlaces(magnitude))}
</span>
```

Money was written as a dollar sign and a number, a duration as its hours, a colon and its minutes, a density's unit as a solidus, a name and an exponent. The browser shapes each run of text on its own, so the same characters written in one string rasterize slightly differently — kerning across what used to be a boundary.

## How different

Measured on the one case that showed up in CI, `R+45.0` on a colorbar:

| | |
|---|---|
| pixels differing | 30, in a single 7×7 box — one glyph |
| largest channel change | 66 of 255, on an antialiasing pixel |
| total absolute difference | 1575, against `check_images.py`'s ignore-threshold of 15 |

Invisible at size; nowhere near invisible to the comparison. 546 of the 1440 cases in the rendering sweep differ this way, and none of them differ in a single character — the sweep renders with `renderToString`, whose `<!-- -->` markers make run boundaries visible, and stripping those makes every pair identical.

**This needs `!updateScreenshots`.** That is the point of doing it first and alone: every later change to how a number reads would otherwise have to reproduce where the pieces happened to fall, and #2229 shows what happens when one doesn't.

## Test plan

- The sweep against `main`: 546 differences, all run-structure, zero textual.


kavigupta:
!updateScreenshots

kavigupta:
!updateScreenshots

PR #2249, opened 2026-08-16
Leave the percent sign off a party quantity we do not have

kavigupta:
Every other unit stopped naming itself for a missing value in #2233, but the party-colored ones kept writing `N/A%`: they render through `PartyPercentage` rather than through `display()`, so the early return that drops the unit never applied to them.

`partyDisplay` now takes the missing case itself, returning `N/A` with a blank unit, which also lets `PartyPercentage` assume it has a value.

The N/A test covers a lead, a share and a change alongside the rest, rather than excusing them.

## Verification

- `tsc`, `eslint .`, `knip`, unit tests
- The rendering sweep against `main`: 42 differences, every one of them a missing value losing its `%`.
- Screenshots will need `!updateScreenshots`.


kavigupta:
!updateScreenshots

PR #2253, opened 2026-08-17
Choose the unit a count is written in by what the number costs to read

kavigupta:
Fourth of the pieces of #2215, off current main.

`population`, `usd` and `fatalities` each had an arm that knew when to abbreviate, sharing one `abbreviate` helper whose ladder and thresholds were stated nowhere. They now say what they are counting:

```ts
number: dimensionfull({}),
population: dimensionfull({ person: 1 }),
fatalities: dimensionfull({ fatality: 1 }),
usd: dimensionfull({ usd: 1 }),
```

and the unit is searched for rather than decided. Each base unit has a ladder of units it can be written in; `bestRepresentation` writes the number out in every one of them and keeps the cheapest, where a digit past the third costs one, a leading zero after the point costs nine tenths, and reaching for an abbreviation costs one.

That reproduces the shared ladder, including the threshold nobody had written down — a thousand is not worth reaching for until the number is five digits long, because `12 345` costs two and `12.3k` costs one — and it reproduces the promotion at `999 500`, because the cost is counted from the number *as printed*, and `999.5k` prints as `1 000k`.

Cost, ladders and `digitCost` are general over products of powers, but nothing uses a power other than one yet. Conventions, the imperial ladders and the `/` in a unit name arrive with area and density.

## Rendering differences

The 1440-case sweep against main is identical. A 60,672-case stress run over the four types, spread across fifteen magnitudes with every value around each of main's thresholds, has **70 differences in three groups**:

| | main | after | when |
|---|---|---|---|
| 66 | `1.00e+3B` | `1 000B` | at or above 999.5 billion |
| 2 | `10 000` | `10.0k` | `9999.5` exactly |
| 2 | `-999 000` | `-1 000 000` | a bare number at an exact negative half |

The first is the `toPrecision(3)` scientific-notation bug — the same one the existing population tests were written to guard against at the lower tiers, which nobody extended to the top of the ladder. A count of a trillion people is not reachable, but a dollar figure is.

The second is the promotion rule applied at the bottom of the ladder: `9999.5` prints as five digits, so it is abbreviated for the same reason `999 500` is. Only reachable for a non-integer count.

The third is a fix I had to make to get the rest right, below.

## Rounding a half by size

`formatToSignificantFigures` rounded with `Math.round`, which takes a half towards positive infinity. So `999 500` read as `1.00m` but `-999 500` read as `-999k`: the ladder promoted one and not the other. It now rounds the magnitude and puts the sign back, which is the rule this stack already applies to which unit and how many places a negative value gets.

This also reaches captions in `derive-human-readable-name.ts`, which share the helper.

## Verification

`tsc`, lint, knip, 530 unit tests. New tests cover the symmetry of a rounded half, the top of the ladder, and `-999 500` reading as `-1.00m`.

Screenshots should not move: every difference above needs a count of a trillion or a fractional one.

### DIFF
--- a/react/src/components/unit-display.tsx
+++ b/react/src/components/unit-display.tsx
@@ -3,7 +3,7 @@ import React, { CSSProperties, ReactNode } from 'react'
 import { useColors } from '../page_template/colors'
 import { HumanReadableElement, reifyReact } from '../utils/human-readable-name'
 import { Hue, missingValue, StoredUnit, writeQuantity } from '../utils/quantity'
-import { abbreviate, formatToSignificantFigures, roundToDigits, separateNumber } from '../utils/text'
+import { roundToDigits, separateNumber } from '../utils/text'
 import { convertPrecipitation, storedUnits, UnitType } from '../utils/unit'
 
 export interface UnitDisplay {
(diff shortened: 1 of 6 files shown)


kavigupta:
Split into #2255, #2256 and #2257 — it was doing three things at once: a rounding fix, a skeleton that changes nothing, and the ladder that carries every rendering difference.

PR #2260, opened 2026-08-18
Label the flag in screenshots "UNIVERSE"

### LINKED ISSUE #2240: Put "Universe" under flag in screenshots for consistency

lukebrody:
Screenshots showed a bare flag, while the header and the embed card both label it. Fixes #2240.

The label follows the embed card's treatment rather than the header's — regular weight at the embed's size and gap relative to the flag, not bold branding purple. Its color is the theme's secondary text color, because the embed's muted gray is picked for that card's fixed cream background and would disappear in a dark-mode screenshot.

The flag and label are centered in the banner as one block, so the flag shifts up slightly instead of the label hanging off the bottom edge.

Every reference screenshot taken from a download on a page with a universe changes, so this needs `!updateScreenshots`.


lukebrody:
!updateScreenshot

lukebrody:
!updateScreenshots

PR #2261, opened 2026-08-18
Pull map drawing apart from the map page

lukebrody:
First of three, ahead of giving mapper links their own embed card. The Worker that draws those cards has no maplibre, no React and no DOM, so this separates the parts of the map page that decide what a map *looks like* from the parts that put it on screen. Nothing changes about what the site does.

`map-rendering` takes over the colours, ramp bins and marker sizes a map result implies, along with the centroid loading behind a point map. `cluster-geometry` takes over the clustering parameters and the pie-slice geometry — the card reproduces the page's clustering with its own supercluster, and if the two read those numbers from different places they silently disagree. `utils/geometry` and `consolidated-shapes` stop making every caller assemble the same loads, and `execute-request` splits the USS interpreter's entry point away from the Web Worker plumbing so a caller without a Worker can use it.

The one change here that isn't strictly behaviour-preserving: a pie chart slice is now one arc rather than four quarter-arcs.

Stack: **01-map-refactors** → [02-map-title](../../tree/02-map-title) → [03-map-embed](../../tree/03-map-embed)


lukebrody:
!updateScreenshots

PR #2265, opened 2026-08-18
Write a length in whichever length reads shortest

kavigupta:
Fifth of the pieces of #2215. Converts `distanceInM`, `distanceInKm` and `area`.

The precision half of this went in separately as #2267, so what is left here is only *which* unit a length is written in.

Elevations were written in meters however tall they were, distances in kilometers however short, and an area chose between four units against thresholds written into its arm. All three are lengths. They now say so:

```ts
distanceInM: dimensionfull({ m: 1 }),
distanceInKm: dimensionfull({ m: 1 }, 1e3),
area: dimensionfull({ m: 2 }, 1e6),
```

and the pool gains the lengths a reader measures in, which is the first thing that depends on the reader — so the pool is now built per reader rather than being a constant.

An acre goes in as a unit of area in its own right. That is what the pool is for: 4046.86 square meters is the square of no length anybody writes, and the model before #2257 could not have expressed it.

## What moves

43 of the 1440 sweep cases, all in the three converted types, and every one of them is a change of unit rather than of precision:

| | main | after |
|---|---|---|
| `distanceInM` 1234 | `1 234 m` | `1.23 km` |
| `distanceInM` 12345, imperial | `40 502 ft` | `7.67 mi` |
| `distanceInM` 0.123 | `0.123 m` | `12.3 cm` |
| `distanceInKm` 0.123 | `0.123 km` | `123 m` |
| `distanceInKm` 0.005, imperial | `0.003 mi` | `16.4 ft` |

**Worth deciding on deliberately: elevations now read in kilometers once they pass a thousand meters.** Denver would be `1.61 km` rather than `1 609 m`. That is what "a length is a length" costs, and it is the one change here I would not want to discover in a screenshot. Say if you would rather elevation stayed in meters — the way to do that is a style for `m^1` that pins the unit, not a special case in the search.

Two units main never showed appear at the extremes: `12.3 cm` for a tenth of a meter, and `129 ft²` for an area of a hundred square feet. Nothing real is that small — the smallest area in the data is a fraction of a square kilometer — but the sweep exercises it.

An area otherwise matches main, including the crossover to acres below a square mile, which now falls out of the cost rather than being a threshold: `0.005 km²` reads `1.24 acres` on both sides.

## Verification

The 1440-case sweep against main: 43 differences, every one in `distanceInM`, `distanceInKm` or `area`, enumerated above. `tsc`, lint, knip, 565 unit tests, including the new lengths and the imperial acre.

Needs `!updateScreenshots`: anything showing an elevation, a mean distance or an area moves.


kavigupta:
!updateScreenshots

PR #2267, opened 2026-08-18
Write a distance to three figures, as everything else of its size is

kavigupta:
Split out of #2265 so that the screenshots it moves can be looked at on their own.

A distance in kilometers was written to a fixed two decimal places and one in meters to none, while every other measured quantity on the site is written to three significant figures. So a mean distance read as `12.30 km` where an area of the same size read as `12.3 km²`, and an elevation of `5.67 m` read as `6 m`.

```diff
 case 'distanceInKm':
-        number: separateNumber((useImperial ? value / kmPerMile : value).toFixed(2)),
+        number: roundToDigits(useImperial ? value / kmPerMile : value, { significantDigits: 3 }),
 case 'distanceInM':
-        number: separateNumber((useImperial ? value * feetPerMeter : value).toFixed(0)),
+        number: roundToDigits(useImperial ? value * feetPerMeter : value, { significantDigits: 3 }),
```

## What moves

62 of the 1440 sweep cases, all in these two types, and **no unit name changes anywhere** — a kilometer stays a kilometer:

| | main | after |
|---|---|---|
| `distanceInKm` 12.3 | `12.30 km` | `12.3 km` |
| `distanceInKm` 1234 | `1 234.00 km` | `1 234 km` |
| `distanceInKm` 0.123 | `0.12 km` | `0.123 km` |
| `distanceInM` 5.67 | `6 m` | `5.67 m` |
| `distanceInM` 12.3, imperial | `40 ft` | `40.4 ft` |

A zero reads `0.000 m` rather than `0 m`, since three figures of nothing is what `roundToDigits` writes when it cannot tell how large a value is. That is already how an area of zero reads on main, so the two now agree.

#2265 keeps the rest: which unit a length is written in, and acres.

## Verification

The 1440-case sweep against main, with every difference listed above by hand. `tsc`, lint, 555 unit tests, including a new block asserting three figures wherever the point falls.

Needs `!updateScreenshots` — every mean distance and elevation on the site gains or loses a digit.


kavigupta:
!updateScreenshots

PR #2278, opened 2026-08-21
Use the exact definitions of the imperial lengths

kavigupta:
Split out of the rainfall change, where the inch first mattered.

The imperial lengths were rounded approximations: a mile as `1609.34` meters and a foot as `1 / 3.28084`. Their exact definitions are a mile of 5280 feet, a foot of twelve inches, and an inch of 2.54 centimeters:

```ts
// the exact definitions: an inch is 2.54cm, a foot twelve of them, a mile 5280 feet
const metersPerInch = 0.0254
const metersPerFoot = 12 * metersPerInch
const metersPerMile = 5280 * metersPerFoot
```

The old constants were out by about two and a half parts in a million, which is enough to show on any quantity written to six digits.

## What moves

Real areas, in their last digits, and the new value is the published one:

| | main | after | published |
|---|---|---|---|
| Alaska | `665 387 mi²` | `665 384 mi²` | 665,384 |
| Texas | `268 598 mi²` | `268 597 mi²` | 268,596 |

Rhode Island and San Francisco do not move — five digits and three are not enough to reach the error.

In the sweep it shows as 7 of 1440 cases, all imperial, all at values with six or more digits: areas, densities per square mile, and distances in miles.

Precipitation is also affected, though nothing renders it yet in these units: the site converted rainfall to inches with an exact `2.54` in one place while deriving the inch from `3.28084` in another, so the two disagreed. They now agree, which is what let the rainfall change land clean.

## Verification

The 1440-case sweep against main, with all 7 differences enumerated above and checked against real areas. `tsc`, lint, 578 unit tests, including two pinning a round mile and a round square mile.

Needs `!updateScreenshots`: any imperial area, distance or density with six digits.


kavigupta:
!updateScreenshots

PR #2280, opened 2026-08-21
Give a quantity its dimensions, a decoration and whether it is a difference

kavigupta:
Stacked on #2279; the diff shown is against it.

`Unit` was five kinds that were really two things:

```ts
export type Unit = (
    { kind: 'temperature' }
    | { kind: 'scalar', dimensions: Dimension[], decoration: Decoration, difference: boolean }
)

export type Decoration = { kind: 'none' } | { kind: 'percent', party?: Party }
export type Party = { kind: 'color', hue: Hue } | { kind: 'lead', system: PartySystem }
```

A temperature stays its own kind because it is the one quantity that is not a scaling of anything — Fahrenheit to Celsius has an offset, and no unit off a ladder can express that. Everything else is a scalar with dimensions, whatever it is dressed in, and whether it is a change rather than a level.

`difference` is where the duplication was. A percentage change differs from a percentage, and each party's change from that party's share, by a plus sign and nothing else — seven kinds for one bit. It is also the version that survives #2216: the difference of two columns is a difference whatever they were measuring, whereas `delta-percentage` only knew about fractions.

`Party` splits the two things the old `partyColor`/`partySystem` pair conflated: a share is always drawn in its party's color, while a margin reads its party off which way it leans, and colors and labels itself accordingly.

The declarations get shorter for it:

```ts
partyPctBlue: percentage(inParty('blue')),
partyChangeBlue: percentage(inParty('blue'), true),
democraticMargin: percentage({ kind: 'lead', system: 'democratic' }),
temperature: { unit: { kind: 'temperature' }, toBaseUnits: 1 },
```

Conventions stay keyed by dimensions in this one. Moving them onto the decoration — so that a statistic keeps its units while a quantity derived from one floats — is the next step, and is what the `decoration` field is there to hold.

## Verification

The 1440-case sweep is identical to #2279, which is itself identical to main. `tsc`, lint, knip, 582 unit tests.

No screenshots should move.

### DIFF
--- a/react/src/utils/quantity.ts
+++ b/react/src/utils/quantity.ts
@@ -19,12 +19,19 @@ export interface Dimension {
 }
 
 // Abstract interpretation of a quantity as a unit.
+/** Whose color a quantity is drawn in, either fixed or read off which way the quantity leans. */
+export type Party = { kind: 'color', hue: Hue } | { kind: 'lead', system: PartySystem }
+
+/** What a quantity is dressed in, over and above the dimensions it has. */
+export type Decoration = { kind: 'none' } | { kind: 'percent', party?: Party }
+
+/**
+ * A temperature is the one quantity that is not a scaling of anything: Fahrenheit to Celsius has
+ * an offset, so no unit off a ladder can express it.
+ */
 export type Unit = (
-    { kind: 'raw-percentage', partyColor?: Hue }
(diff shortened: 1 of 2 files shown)


kavigupta:
Rebuilt on main as the restructure now lands first; superseded by the new PR.

PR #2297, opened 2026-08-22
Multiply, divide and raise units, and say what an expression is in

kavigupta:
Off main. The first two pieces of #2216, the second having been merged in here as #2299.

## Arithmetic on units

Inferring the unit of a computed quantity means saying what the unit of a product is, of a quotient, and of a quantity raised to a power. That is arithmetic on the dimensions, with what each is stored in carried along:

```ts
export function unitProduct(left: StoredUnit, right: StoredUnit, rightPower: 1 | -1): StoredUnit | undefined
export function unitPower(stored: StoredUnit, exponent: number): StoredUnit | undefined
export function sameDimensions(left: StoredUnit, right: StoredUnit): boolean
```

A kilometer squared is a million square meters, so the scale is raised with the dimensions, and dimensions that cancel are dropped rather than left at zero. Nothing computed keeps what its statistic was written in: a population over an area is people per square meter, to be written in whatever reads shortest, where the density *statistic* is per square kilometer because it says so.

Two temperatures are the same kind of thing, so one can be taken from the other. Twice a temperature is another matter — how many temperatures it is depends on the two of them rather than on their units — so a product or a power involving one is no quantity at all.

## What an expression is in, and what its operands must be

Inference needs both directions. Forwards gives the unit of a computed column; backwards reads the `0.1` of `commute_bike < 0.1` as ten percent, which is what the captions need.

```ts
export type AbstractInterpValue = (
    { kind: 'any', constant?: number }
    | { kind: 'none' }
    | { kind: 'in', unit: StoredUnit, constant?: number }
)
```

Not knowing which unit something is in is a different thing from nothing being of any unit, and only the second is worth telling a script's author about. Both are ordinary members of the domain, so every expression has an answer and no operation can fail — a test runs every operator over every kind of value in all three directions and asserts none of them throws.

Every operator is one of three shapes rather than a case of its own: the slots of a **sum** are all the same unit, the slots of a **product** have dimensions that add, and a **power** needs its exponent to be a constant. Solving for an operand is running the operator that undoes it —

```ts
const inverses = { '+': '-', '-': '+', '*': '/', '/': '*' } as const
```

— so each direction is the other read backwards rather than a second set of rules to keep in step. The unary operators are their own inverses, and say so.

Constants are in the domain because coefficients need them. A bare number is dimensionless under *product* and takes the other side's unit under *same unit*, which is why `pop * 2` and `pop + 2` differ; and negating one negates it, so `pop * -2` scales by minus two.

Coefficients are checked where the quantity is written, not as it goes:

```
t1 - t2            times=0     a number of degrees
t1 + t2            times=2     carried, but nothing writes it
(t1 + t2) / 2      times=1     a temperature again
t + 2              times=1     a temperature, the 2 being degrees
-t                 times=-1    no reading: -50F and -10C are not the same one
```

Refusing the sum on the way past would have lost the mean, which is why `unitToWriteIn` asks at the end.

## Verification

`tsc`, lint, knip, 615 unit tests, twenty-four of them new. The 1440-case rendering sweep is unchanged: nothing here is reachable from a statistic yet.

Next is the walk over the AST, then the two entry points, then the captions.

### DIFF
--- a/react/src/urban-stats-script/unit-algebra.ts
+++ b/react/src/urban-stats-script/unit-algebra.ts
@@ -1,0 +1,192 @@
+import { assert } from '../utils/defensive'
+import { dimensionless, sameDimensions, StoredUnit, unitPower, unitProduct } from '../utils/quantity'
+
+import { BinaryOperatorSymbol, UnaryOperatorSymbol } from './operators'
+
+/** `any` is that it could be in any unit; `none` is that no quantity is of it, as people plus an area. */
+export type AbstractInterpValue = (
+    { kind: 'any', constant?: number }
+    | { kind: 'none' }
+    | { kind: 'in', unit: StoredUnit, constant?: number }
+)
+
+export function constant(value: number): AbstractInterpValue {
+    return { kind: 'any', constant: value }
+}
+
+export function inUnit(unit: StoredUnit): AbstractInterpValue {
+    return { kind: 'in', unit }
+}
+
+/**
+ * Asked at the end rather than at every step: two temperatures added are neither one temperature
+ * nor none, but their mean is one again, and refusing the sum on the way past would lose that.
+ */
+export function unitToWriteIn(known: AbstractInterpValue): StoredUnit | undefined {
+    if (known.kind !== 'in') {
+        return undefined
+    }
+    const { unit } = known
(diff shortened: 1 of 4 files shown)


kavigupta:
Split back apart: the dimension arithmetic is #2300, off main, and the abstract interpretation is #2301 on top of it. #2299 having merged in here had run the two together.

PR #2305, opened 2026-08-23
use unit inference for maps and tables

kavigupta:
Towards #2204. The first piece of it that reaches the screen.

## What a map and a column do today

Both state their units or go without, and going without, they are written in whatever a **string match on the derived name** takes them for:

```ts
const unit = ramp.unit ?? unitTypeToStoredUnit(classifyStatistic(reifyString(ramp.label)))
```

`classifyStatistic` looks for `Density`, `Population`, `%`, `Fatalities` in the name. So a map of `population / area` is labelled `Population ÷ Area`, matches `Population`, and is written **in people**. A map of `density_pw_1km / density_pw_2km` has `Density` in its name twice and is written **per square kilometre**, though it is a bare ratio.

## What lands

`deriveMapUnit` / `deriveTableColumnUnit`, which run the inference the last few PRs built over the `data=` or `values=` expression, and are consulted exactly where the unit is not stated. The name match stays as the fallback for anything the script cannot be read for — a call, an attribute, an unknown identifier.

The script is read **whole**, so a map of a name it assigned is in the units of what was assigned:

```
x = population / area
condition (true)
cMap(data=x, ...)          →  people per km²
```

Everything I compared against the name match, at a stored value of 1 234:

| script | today | now |
|---|---|---|
| `population / area` | `1 234` | `1 234 / km²` |
| `traffic_fatalities / area` | `1 234` | `1 234 / km²` |
| `density_pw_1km / density_pw_2km` | `1 234 / km²` | `1 230` |
| `high_temp - low_temp` | `1 234.0°F` | `+1 234.0°F` |
| `population - population_2000` | `1 234` | `+1 234` |
| `density_pw_1km`, `population`, `area`, `elevation`, `commute_bike`, `high_temp` | unchanged | unchanged |

## Refusing what has no writing

Two failures the comparison turned up, both now refused so the name match keeps the case:

- **`population ** 0.5` threw.** No pool holds a unit for half a person, so the search found no covering and hit `assert(best !== undefined, 'every quantity has at least one way of being written')`. That assertion is false of fractional dimensions, and only a derived unit can have them.
- **A count vanishes where it is not the thing being counted.** People go unnamed — that is why a density reads `/ km²` — so `population * area` was written `1.23 m²·B` and `area / population` was written `1 234 km²`, which reads as an area. `writableDimensions` now allows a count only as the single thing above the line.

## Screenshots

I expect movement only where a script's `data=`/`values=` is a derived expression *and* the name match disagreed. Plain-statistic maps and columns are unchanged, which is most of the e2e suite. Happy to `!updateScreenshots` once CI has run.

## Still floating

In imperial, a derived density picks `20.2 / acres` where the density statistic says `12 950 / mi²` — the search reads acres as cheaper, and only declared statistics pin their units. Both are readable, but two maps of the same thing disagree. Say the word and I will pin the dimension rather than the statistic.

## Verification

`tsc`, lint, knip, the full unit suite, 6 new tests in `derive-unit.test.ts` covering the table above and both refusals.


kavigupta:
! updateScreenshots

kavigupta:
!updateScreenshots

PR #2314, opened 2026-08-24
Compare search button

### LINKED ISSUE #2212: button in search to compare
on the right here it should have a button that's like Compare or sth that lets you compare to this region, when you're searching in the top bar

Image

yes, redundant with the comparison bar but i accidentally go to the wrong one all the time

lukebrody:
Closes #2212


lukebrody:
!updateScreenshots

PR #2317, opened 2026-08-24
Make the delete manipulation button a link

lukebrody:
Deleting a column from a comparison is a navigation, so the control is now an anchor carrying the resulting comparison URL. As a `<button>` it hit the early return in `Navigator.link` on ctrl/cmd/shift-click and did nothing at all; as an anchor the browser handles modifier-clicks, middle click, and the context menu's open-in-new-tab and copy-link-address, which JS emulation cannot match faithfully across browsers.

Replace stays a `<button>` because it toggles a search box rather than navigating, so `ManipulationButton` takes either an `onClick` or a `NavLink` and renders the matching element. `withButtonRole` supplies the button role and Space activation on the anchor, and the `a[role=button]` rules from the base of the stack give it the box metrics an anchor doesn't inherit.

`HeadingDisplay`'s `includeDelete` and `onDelete` collapse into a single optional `deleteLink`, since there is no meaningful delete URL when only one article is left.

Stacked on #2313.


lukebrody:
!updateScreenshots

lukebrody:
!updateScreenshots

PR #2320, opened 2026-08-24
Write the numbers in a caption in the units they are read from

kavigupta:
Closes the last piece of #2204. `backward` and `backwardUnary` have been exported and tested since #2301 with nothing calling them; this is what they were for.

## What changes

A caption writes the numbers a script names in the units the script reads them from:

| | before | after |
|---|---|---|
| `condition (commute_bike > 0.05 & high_temp > 80)` | `Commute Bike % > 0.05 and Mean high temp > 80` | `Commute Bike % > 5% and Mean high temp > 80°F` |
| `condition (area > 100)` | `Area > 100` | `Area > 100km²` |
| `condition (ped_cyclist_fatalities_per_capita > 1e-5)` | `> 1x10^{-5}` | `> 1/ 100k` |
| `condition (population > 1m)` | `Population > 1m` | unchanged |

0.05 is the fraction the column is stored as, and no reader is reading fractions.

## How

`inferConstantUnits` runs the walk forward and then pushes the result back down the expression, recording what each number was wanted as. A comparison is what usually carries it: `backward('<', ⊤, commute_bike, 'right')` answers "of the same kind as the other side".

What a script says nothing about stays a bare number, which is most numbers:

```
population * 2       the 2 scales a quantity and is none
population / 1000    likewise
high_temp - 32       left alone, see below
maximum(population, 1000)   1 000, of the kind of the other side
```

## One fix to `backward`

A sum whose total is unknown now gives an operand the other's dimensions **without** saying how many of it. A temperature and a difference of two add to each other either way round, so `high_temp - 32` cannot say whether the 32 is a reading or a number of degrees — and writing `32°F` would be wrong in Celsius, where it would read `0°C` rather than the 17.8 degrees the script meant. Before this it claimed a reading.

## Not in this PR: the reader's own units

Captions are derived where the reader's settings are not to hand, so a quantity is written metric and Fahrenheit regardless. An imperial reader sees `100km²` in the caption above a colorbar reading `mi²`. Fixing it means either threading `ReaderSettings` into the derivers or carrying a quantity element in `HumanReadableName` and formatting it at display — the latter is blocked on `quantity.ts` and `human-readable-name.tsx` importing each other. Worth its own change; today's captions say `0.05`, so this is strictly better either way.

## Verification

`tsc`, lint, knip, the full unit suite with 12 new cases, and 10 e2e tests — one new, asserting the page title of a filtered table reads `Population where Commute Bike % > 5% and Mean high temp > 80°F`.

One existing label expectation moved, from `1x10^{-5}` to `1/ 100k`.


kavigupta:
!updateScreenshots

kavigupta:
!updateScreenshots

kavigupta:
!updateScreenshots

PR #2323, opened 2026-08-24
Write a caption's numbers in the units of whoever is reading

kavigupta:
Stacks on #2320, which is where a caption's numbers first get units at all. **Do not merge before it.** Unblocked by #2321.

## The gap #2320 left

A caption is derived where the reader's settings are not to hand, so its numbers were written in whichever units the site is written in. An imperial reader got a colorbar in `mi²` under a caption saying `100km²`.

## The fix

A number in a name is carried as the quantity it is, rather than as text decided when the name was derived:

```ts
| { type: 'quantity', value: number, unit: StoredUnit }
```

and written where the name is written. `reifyReact` takes the reader's settings, so a page asks whoever is reading; `reifyString` and the embed Worker's own walker write it the way they always did, there being nobody to ask in a title, a card or a downloaded file.

Three components pass what they read in — the colorbar's label, a statistic panel's subtitle, and a column's name — through one `useReaderSettings` hook, the same two settings `Statistic` already reads for the values below.

```
max(Mean high temp, 80°F)      →  max(Mean high temp, 26.7°C)
min(Area, 100km²)              →  min(Area, 38.6mi²)
max(PW Density (r=1km), 5 000/ km²) → max(PW Density (r=1km), 12 950/ mi²)
```

## Why the settings are passed rather than hooked

`reifyReact` could have read the settings itself, but the embed Worker imports `reifyString` from the same module, so a hook there would drag the settings store and its data imports into the Worker's bundle. An argument with a default costs three call sites instead.

## Verification

`tsc`, lint, knip, the full unit suite with four new cases, and 11 e2e tests — one new, which reads a column's name as `max(Mean high temp, 80°F)`, switches the reader to Celsius, and reads it again as `max(Mean high temp, 26.7°C)`.


kavigupta:
Closed in favour of #2346, which is the same change rebuilt on today's main — this branch predated the caption work landing (#2320), the placement API (#2345) and the one-run rendering, and every file in it conflicted with a later rewrite of the same lines.

PR #2328, opened 2026-08-25
Edit a custom table beside its results, the way a map is edited

lukebrody:
Stacked on #2327, which moves the layout this uses out of the mapper. #2335 sits on top of this one with an unrelated button restyle.

Editing a custom table pushed the results below a tall settings panel, so a change and its effect could not be seen at once. Editing now uses the same split view the mapper has: settings on the left, and the table on the right, along with its title and its View / Convert to Map buttons. Viewing is untouched, and so is editing on mobile, where there is no room for two columns.

The sidebar collapses into the hamburger while editing, which is what frees the width for two columns. That decision now reads the page descriptor rather than the page data, because the statistic panel toggles edit mode by updating the descriptor in place and never reloads the page — the page data it loaded with goes stale the moment the user clicks Edit.

Three existing tests broke on the narrower, self-scrolling settings pane, each for a reason worth knowing. Clicking a code editor put the caret wherever the click landed, which was reliably past the end of the text only because the editor used to be wide, so `typeInEditor` now starts from the end of the content. Dragging an AutoUX row onto one further down needs the target scrolled into the pane first, exactly as the mapper's own reorder test already does. And one test reached for the Name and Unit checkboxes through the sidebar helper, which was a no-op on this page until the sidebar started collapsing — with a hamburger present it opened the sidebar over the checkboxes and swallowed the clicks.

Two new tests cover what previously had only screenshots behind it: that the split appears in edit mode and not in view mode, that the sidebar collapses and comes back, that the title and buttons sit on the table side, and that mobile still stacks. The mobile one is the statistic page's first coverage at that width.

The reference screenshots for the edit-mode statistic tests need regenerating through `!updateScreenshots`.


react/src/page_template/utils.ts
```diff
@@ -1,22 +1,25 @@
...
+            return page.edit === true
         default:
             return false
     }
 }
 
 export function useHideSidebarDesktop(): boolean {
     const navigator = useContext(Navigator.Context)
-    return hideSidebarDesktop(navigator.usePageState().current.data)
+    // The descriptor rather than the page data, since the statistic panel updates it in place as edit mode is toggled
```
lukebrody:
```suggestion
```

react/src/stat/StatisticPanelPage.tsx
```diff
@@ -85,6 +94,46 @@ export function StatisticPanelPage({ view, stat, data, set, loading, counts, err
...
+    )
+
+    return (
+        <PageTemplate
+            showFooter={!splitLayout}
+            screencap={data && ((...args) => createScreenshot(() => ({
+                path: `${sanitize(reifyString(data.renderedStatname))}.png`,
+                overallWidth: tableRef.current!.offsetWidth * 2,
+                // Resolved inside screenshot mode: the disclaimer has moved into the footer by
+                // now, so include the footer as a footnote below the table when it has content.
```
lukebrody:
```suggestion
```

react/test/statistic-column-reorder.test.ts
```diff
@@ -122,6 +122,8 @@ test('reordering via autoux', async (t) => {
     await waitForLoading()
     await t.click('button[data-test-id="edit"]')
     await toggleCustomScript(t)
+    // the settings pane scrolls on its own, so the target has to be brought into it first
```
lukebrody:
```suggestion
```

react/test/statistics.test.ts
```diff
@@ -406,6 +406,45 @@ test('edit starting from a statname page works', async (t) => {
...
+
+test('editing on desktop puts the settings beside the table', async (t) => {
+    await t.expect(splitLeft.exists).notOk()
+    await t.expect(hamburgerMenu.exists).notOk()
+
+    await t.click(Selector('button[data-test-id="edit"]'))
+    await waitForLoading()
+
+    await t.expect(splitLeft.exists).ok()
+    // the sidebar collapses into the hamburger, as it does on the mapper
```
lukebrody:
```suggestion
```

lukebrody:
!updateScreenshots

lukebrody:
!updateScreenshots

lukebrody:
!updateScreenshots

PR #2332, opened 2026-08-25
Edit a mapper filter graphically instead of only as code

lukebrody:
Stacked on #2331.

A filter was a code box and nothing else, so writing one meant knowing USS. Most filters are the same shape — some statistics compared against thresholds, combined with and/or — so that shape gets a UI, and anything outside it still falls back to the code box.

The grammar is a view over the ordinary AST rather than a parallel representation. A group is a chain of `&` or `|` operands, flattened so `a & b & c` reads as three rows rather than nested pairs, and a comparison is a `number[]` against a constant or another `number[]`. Two things follow from staying on the real AST: scripts remain infix and hand-editable, and operator precedence gives nesting for free — `a & b | c` opens as an "Any of" containing an "All of" without any special handling. Conditions saved before this change are read back into the grammar rather than left as code, so existing maps come up graphically; leaving the code box re-parses what you typed rather than keeping it as one opaque operand.

Two latent bugs surfaced on the way, both fixed here because this is the first branch where they are reachable:

- `unparse` treated a custom node as atomic. That was harmless while custom nodes only ever sat in comma-delimited argument lists, but it would have dropped the parentheses around a custom `a | b` inside an `&` and silently changed what the script meant. Wrappers are now resolved to what they actually print — including the auto-UX rewrite, which is why a typed constant no longer exports as `x > (5000)`.
- A selector's `<input>` carried `min-width: auto`, so it kept its ~147px intrinsic width and overlapped whatever shared its row. Invisible until the operator dropdown put something beside it.

The export/import fixture changes because a filter now round-trips as `condition (population > 10000)` rather than `condition (customNode("population > 10000"))`. Three statistics tests that typed into the filter's code box now select "Custom Expression" first, since checking "Filter?" gives a comparison.

16 unit tests cover the grammar — parsing, flattening, precedence nesting, custom fallback at both node and operand level, kind switching, block-ident consistency, and a save/reload round trip. 8 e2e tests drive the UI. Screenshots will need regenerating.


lukebrody:
!updateScreenshots

react/src/mapper/settings/BetterSelector.tsx
```diff
@@ -144,8 +144,9 @@ export function BetterSelector<T>({ value, onChange, possibleValues, renderValue
     // eslint-disable-next-line no-restricted-syntax -- Dynamic tag name
     const InputElem = iframe ? IFrameInput : 'input'
 
+    // `auto` rather than a zero basis so a squeeze is shared with whatever is beside this, not taken entirely out of it
```
lukebrody:
```suggestion
```

react/src/mapper/settings/BetterSelector.tsx
```diff
@@ -177,6 +178,8 @@ export function BetterSelector<T>({ value, onChange, possibleValues, renderValue
                 placeholder="Search options..."
                 style={{
                     flex: 1,
+                    // Without this the input keeps its intrinsic width and overruns whatever is beside it
```
lukebrody:
```suggestion
```

react/src/mapper/settings/condition.ts
```diff
@@ -0,0 +1,204 @@
+/**
+ * The subset of USS that the condition editor can display graphically:
+ *
+ *     condition = condition & condition | condition | condition
```
lukebrody:
```suggestion
 *     condition = condition & condition | (condition | condition)
```

react/src/mapper/settings/condition.ts
```diff
@@ -0,0 +1,204 @@
...
+import { changeBlockId, createDefaultExpression, parseExpr } from './parseExpr'
+
+const groupOperators = ['&', '|'] as const
+export type GroupOperator = typeof groupOperators[number]
+
+export const comparisonOperators = ['==', '!=', '<', '<=', '>', '>='] as const
+export type ComparisonOperator = typeof comparisonOperators[number]
+
+export type ConditionKind = GroupOperator | 'comparison' | 'custom'
+export const conditionKinds = [...groupOperators, 'comparison', 'custom'] as const satisfies ConditionKind[]
```
lukebrody: should probably do the typeof arr[number] here so nothing can be left out of the array

react/src/mapper/settings/condition.ts
```diff
@@ -0,0 +1,204 @@
...
+}
+
+function asCustomCondition(expr: UrbanStatsASTExpression, blockIdent: string): UrbanStatsASTExpression & { type: 'customNode' } {
+    return parseNoErrorAsCustomNode(unparse(expr, { simplify: 'auto-ux' }), blockIdent, conditionTypes)
+}
+
+/**
+ * Normalizes an arbitrary expression into the condition grammar, assigning block idents.
+ * Custom nodes are looked inside of, so that conditions written as code before there was a
+ * graphical editor come up graphically.
```
lukebrody:
```suggestion
```

react/src/mapper/settings/condition.ts
```diff
@@ -0,0 +1,204 @@
...
+        }
+        default:
+            return undefined
+    }
+}
+
+export function changeConditionKind(
+    current: UrbanStatsASTExpression,
+    classified: Condition,
+    kind: ConditionKind,
```
lukebrody: improve the names of these

react/src/mapper/settings/condition.ts
```diff
@@ -0,0 +1,204 @@
...
+    blockIdent: string,
+    typeEnvironment: TypeEnvironment,
+): UrbanStatsASTExpression {
+    if (kind === classified.kind) {
+        return current
+    }
+    if (classified.kind === 'custom') {
+        // Leaving custom is a request to read the code as a condition, not to keep it as one opaque operand
```
lukebrody:
```suggestion
```

react/src/urban-stats-script/parser.ts
```diff
@@ -841,8 +841,23 @@ export function unparse(node: UrbanStatsASTStatement | UrbanStatsASTExpression,
...
+        if (opts.simplify === undefined) {
+            return expr
+        }
+        if (expr.type === 'autoUXNode') {
+            return effectiveExpression(expr.expr)
+        }
+        if (expr.type === 'customNode' && expr.expr.type === 'expression') {
+            return effectiveExpression(expr.expr.value)
+        }
+        // Rules are applied to whatever node `unparse` is handed, so the printed form can be simpler than the node
```
lukebrody: These comments aren't great. Improve or remove

react/test/mapper-condition.test.ts
```diff
@@ -0,0 +1,120 @@
...
+
+mapper(() => test)('unchecking the filter clears it', { code: base }, async (t) => {
+    await toggleCustomScript(t)
+    await checkTextboxesDirect(t, ['Filter?'])
+    await replaceInput(t, '0', '10')
+    await checkTextboxesDirect(t, ['Filter?'])
+    await t.expect(getInput('Comparison').exists).notOk()
+    await toggleCustomScript(t)
+    await t.expect(getCodeFromMainField()).notContains('condition')
+})
```
lukebrody, line 123: We should test error handling/routing

PR #2333, opened 2026-08-25
Line a list row's controls up with the row's selector

lukebrody:
Stacked on #2332.

A list element's remove button and drag handle were centred on the whole element, which is correct only when the element is one line tall. A nested call is several lines tall, so the controls drifted down beside the element's arguments instead of sitting next to the thing they remove.

Centring them on the selector alone requires them to share a flex line with it — but that widens everything underneath to run past the button. The two constraints can't both be met by a flex row, since the button is either a sibling of the selector (right centre, wrong width) or a sibling of the whole element (right width, wrong centre). A two-column grid does both: header and controls share row one, and the body is pinned to the header's column. The drag handle stays absolutely positioned outside the row, anchored to the header rather than the element, so list rows still line up with non-list editors.

Column one is `minmax(0, 1fr)` rather than `1fr`, because a bare `1fr` track takes an automatic minimum from its content and the 200px constant input inside was pushing rows wider than the settings panel.

Measured in the browser rather than eyeballed: for a custom ramp, each element's header and body both span `89.5..516.1` with the button at `524.1..548.0`, centred within half a pixel of the selector.


lukebrody:
!updateScreenshots

PR #2335, opened 2026-08-25
Let the statistic page's buttons look like every other button

lukebrody:
Stacked on #2328.

View, Filter / Edit Table, and Convert to Map each carried a hand-written copy of the button styling — background, text color, border, radius, cursor — that differed from what `common.css` gives every other button on the site, most visibly in the border color and in having no hover state. They now inherit it and keep only the layout each one needs.

This is a view-mode visual change, which is why it is its own PR rather than part of the split-view work: it moves the reference screenshots for the statistic page anywhere one appears, including in `article-universe`, `cross_source_border_disclaimer`, and `navigation`. Those need `!updateScreenshots`.


lukebrody:
!updateScreenshots

PR #2343, opened 2026-08-26
Photograph the settings pane from the top rather than from wherever a click left it

lukebrody:
A pane that scrolls inside itself puts its scroll offset into every screenshot, and nothing was pinning that offset — it was wherever the last click had scrolled to. That position moves whenever content above it grows asynchronously or an element changes height, which is what has been shifting the settings pane between otherwise identical runs and forcing repeated reference regeneration. The same shift is why a click computed just before one can land on the wrong control.

Screenshots now reset the pane to the top first, so the framing is a property of the shot rather than of the interactions preceding it. Three shots whose subject lives below that fold name it with the new `scrollPaneTo` option — the second field's error and the warning on the table page, and the mapper's Insets checkbox — which scrolls back to it from a known starting point rather than inheriting one.

Two consecutive local runs of `statistics.test.ts` now agree to within antialiasing: the worst image differs by 31 pixels and none exceeds 100, where the pane previously moved wholesale. `statistics.test.ts` (58) and `mapper-ux_x0.test.ts` (11) pass.

The mapper's pane has always had this drift, so the reset moves three of its references too. I checked each: every subject survives, with `cluster map enable insets` needing the `scrollPaneTo` to keep its checkbox in frame. `!updateScreenshots` is needed for both the statistic and mapper references.


lukebrody:
!updateScreenshots

PR #2345, opened 2026-08-27
Clean up legend captions and quiz displays to fix spacing issues

kavigupta:
Off main.

## The rule as it was

A unit written *per* something, with nothing in front of the solidus, was always set with a space:

```ts
atom(over.length === 0 ? '/ ' : '/')
```

That is right in a table, where the unit sits in a column of its own and `/ km²` reads as "per square kilometre" beside a number two columns away. It is wrong everywhere else: a legend and a caption write the unit hard against its number, and `5 000/ km²` reads as two things rather than one.

## Now

The space is asked for rather than assumed — `nameOf(written, alone)`, threaded through `writeQuantity` and `renderQuantity` to a `unitAlone` prop on `Statistic`.

| | |
|---|---|
| the table's unit column | asks for it — unchanged |
| the embed card | asks for it: it sets the unit apart from its number with an explicit gap, mimicking that column |
| a map's legend | no longer gets it: `5 000/km²` |
| a quiz's value | no longer gets it |
| a caption (#2320) | no longer gets it |

## What moves

Only quantities written *per* something with a nameless count above the line — density, fatalities per capita, an inverted area. Screenshots of map legends and quiz values with such units will move by the width of a space; the table and the card will not.

## Verification

`tsc`, lint, knip, the full unit suite. 28 expectations across three test files lose the space, being inline renderings, and a new case pins both forms:

```ts
assert.equal(reifyString(nameOf([...perArea], true)), '/ km^{2}')
assert.equal(reifyString(nameOf([...perArea])), '/km^{2}')
```


kavigupta:
!updateScreenshots

PR #2373, opened 2026-08-29
Write a number spelled toNumber("1000") as the number

kavigupta:
The reported map's condition is `density_pw_1km > (toNumber("1000"))`, and its caption read:

> 100% − Commute Car % where PW Density (r=1km) > **Anything to Number**(1000)

`toNumber` is now never written in a caption. It says the value is read as a number, which is what the caption is already doing by writing it:

| condition | caption |
|---|---|
| `density_pw_1km > toNumber("1000")` | `PW Density (r=1km) > 1 000/km²` |
| `high_temp > toNumber("80")` | `Mean high temp > 80°F` |
| `density_pw_1km > toNumber(geoName)` | `PW Density (r=1km) > Default Universe Geography Names [in /km²]` |
| `population > toNumber(geoName)` | `Population > Default Universe Geography Names` |
| `density_pw_1km > -toNumber(geoName + geoName)` | `PW Density (r=1km) > -(… + …) [in /km²]` |

An argument that parses as a literal is written as the number it works out to, in the unit read at that point — the backward pass records the unit at such a call the way it already records one at a literal. Any other argument is written in the call's place, bracketed if it is an operator, since whatever encloses the call decides on brackets by looking at a call. The unit it is read in follows, unless the unit has no name of its own: a count is named by the statistic counting it.

That suffix is an element rather than text, so it follows the reader like the numbers do: the same caption reads `[in °F]` or `[in °C]`, `[in km²]` or `[in acres]`.

The call is matched with `literal-parser`, as the rest of the codebase matches AST shapes. `l.tryParse` is new: four callers were each writing the same catch by hand.

Depends on #2374, which extracts the `asNumber` this exports so `toNumber` and the caption cannot disagree about what parses.

`tsc`, eslint, 770 unit tests, `statistic-inferred-units` and `embed_worker` pass.


kavigupta:
!updateScreenshots

PR #2380, opened 2026-08-29
Lay a taller-than-wide map out at 1200 tall

### LINKED ISSUE #2259: Fix mapper layout for tall geographies
E.g. Chile

lukebrody:
Fixes #2259.

A map's canonical layout size was 1200px wide whatever its shape, so a tall geography got a layout thousands of pixels high — Chile's aspect ratio of 0.18 gave it a 1200x6840 layout. Anything sized in canonical pixels (marker radii above all, but also outline weights and text box fonts) is then a vanishing fraction of the map, which is what the issue is showing.

`canonicalWidth` now takes the map's aspect ratio and returns `1200 * min(1, aspect)`, so neither dimension of a layout exceeds 1200 and a narrow map gets the same pixel budget along its long axis as a wide one. Scaling the long axis rather than the short one is the choice here: it keeps a wide map's rendering bit-for-bit identical, so no existing behaviour moves.

Both callers already had the aspect ratio to hand — the page computes it for `MapLayout`, the embed card computes it to fit the map in its box — so this is a parameter rather than new plumbing. `TransformConstantWidth` takes the derived width as a prop instead of importing the constant.

The new e2e test exports Argentina, whose aspect ratio of 0.45 puts the layout at 545x1200. No existing reference screenshot moves: every map the suite draws is USA or India, and both are at least as wide as they are tall.


lukebrody:
!updateScreenshots

PR #2382, opened 2026-08-29
Warn when a table or a map is drawn in no units at all

kavigupta:
A column with no `unit=` whose unit cannot be worked out, and a map in the same position, quietly fall back to writing plain numbers. Both now warn, naming the part of the script the units come apart in and what went wrong, and carrying that expression's location so the editor underlines it:

> Could not compute units for population + area: cannot add people and m^2 &nbsp;&nbsp;*at 1:11-27*

| script | warning |
|---|---|
| `population + area` | Could not compute units for population + area: cannot add people and m^2 |
| `population * area` | Could not compute units for population * area: people·m^2 has no name of its own |
| `sum(high_temp)` | Could not compute units for sum(high_temp): it is neither a reading nor a difference of readings |
| `geoCentroid.lat` | Could not compute units for geoCentroid.lat: no unit is known for it |

It reports the smallest failing part, so `(population + area) / 2` points at the addition and `high_temp + population * area` at the product. A part whose unit is merely unknown is passed over in favour of one that fails for a reason.

Each warning is raised where the fallback happens — `statDataFromTable` for a column, `computeRampToDisplay` for a map — and travels through the channel that already reports a name that could not be derived. A map warns only where a colorbar is drawn, so `cMapRGB` says nothing.

**A root of a count is now written rather than refused.** A count has no name, so half of one has none either: the covering search may take a nameless unit of size one to any power, and `writableDimensions` allows a count to any positive power. `population ** 0.5` reads `4 825` where it read `4 830`, which was three significant figures of a quantity in no units at all. Two e2e expectations move with it, and it is easily split out if you would rather.

Where each is seen: the mapper shows warnings in its results panel, as it does deprecations. A statistic page shows them while its script is being edited — `StatisticPanelPage.tsx:88` filters a reading view down to errors — so this reaches the person writing the table rather than someone reading it.

Three e2e tests: a map with no derivable unit warns with the location, one with a derivable unit stays silent, and a column reports it in edit mode. `tsc`, eslint, 781 unit tests, `statistic-inferred-units` (15) and the two mapper tests pass. Any existing mapper test whose script has no derivable unit will now show a warning in its panel; those screenshots compare only in CI.


kavigupta:
Superceded by #2393

PR #2388, opened 2026-08-30
Lay the quiz results out as one grid

### LINKED ISSUE #2346: Quiz results table is kinda a mess
It doesn't seem properly aligned, it is very cramped on the values (especially on mobile), etc

lukebrody:
Closes #2346.

The results list was misaligned because each question rendered its own `<table>`. Every row sized its columns independently, so the names, the numbers, and the comparison symbol landed at a different x on every row. One grid for the whole list gives all the rows the same five columns, which is what makes the list read as a table at all.

Two other things were making it look cramped. The values broke between the number and its unit — `QuantityTogether` puts a zero-width space there, and the value column was 10% wide, so the break was always taken; they are `nowrap` now and the columns size to their content. And the question label was `display: table`, so a long one like "Employed in Other services, except public administration %" did not wrap and scrolled the whole page sideways.

Cell padding is `min(0.75em, 2vw)` rather than a flat value. The full amount applies above ~750px; below that the middle group would otherwise grow ~60px on a 524px screen and take it out of the region names, which are already the tightest thing on the row.

The correct/incorrect emoji column is gone. The row carries a faint tint in the result color and the answer you picked carries a stronger one, so choice and correctness are both still encoded, in the space the row already occupies rather than in a sixth column.

The `<a>` wrapping each juxtastat row is `display: contents` so its cells participate in the grid; row clicks still navigate to the comparison page.

The quiz reference screenshots will all need regenerating.


lukebrody:
!updateScreenshots

lukebrody:
!updateScreenshots

lukebrody:
!updateScreenshots

PR #2390, opened 2026-08-30
exact unit inference (try 1)

kavigupta:
Split off #2382, which is about warnings and needs both of these.

**Two quantities add only where they are stored the same way.** The values a script adds are the ones the statistics are stored as, so `hospital_mean_dist + elevation` adds a kilometre number to a metre number and means nothing, though both are lengths. `addedForward` now compares stored scales as well as dimensions, which the rule for the arms of an `if` already did.

**A literal is read in whatever unit the backward pass gave it.** `area + population * 1` is an area: the `1` is a number of square kilometres per person, which is what the caption has written it as all along — the forward pass just disagreed. `Scope` carries the constants the backward pass worked out, and `deriveMapUnit` and `deriveTableColumnUnit` read against them.

That makes `* 1` an escape hatch for a script whose units do not otherwise line up, which is deliberate. It balances the units by assumption rather than by changing the arithmetic, so a caption can show a factor the script does not apply — `hospital_mean_dist > elevation * 1` writes the 1 as `1 000`.

Where nothing can be read into the literal it is still nothing: the `1` inside `ln(population * 1)` cannot make a logarithm into an area, a logarithm making a plain number of whatever it is given.

`tsc`, eslint, 785 unit tests including both rules, `statistic-inferred-units` (14) and `mapper-condition` (10) pass.

### DIFF
--- a/react/src/urban-stats-script/derive-unit.ts
+++ b/react/src/urban-stats-script/derive-unit.ts
@@ -4,14 +4,19 @@ import { StoredUnit } from '../utils/quantity'
 import { UrbanStatsASTExpression } from './ast'
 import { TypeEnvironment } from './types-values'
 import { unitToWriteIn } from './unit-algebra'
-import { inferBindings, inferUnit } from './unit-inference'
+import { inferBindings, inferConstantUnits, inferUnit } from './unit-inference'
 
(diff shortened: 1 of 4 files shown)


kavigupta:
Superceded by #2393

PR #2394, opened 2026-08-30
Show USS warnings as warnings, in the right place

### LINKED ISSUE #2389: warnings should have a yellow not red underline in uss

lukebrody:
Two display bugs made a deprecation warning in a mapper script look like a broken script.

A deprecated identifier was underlined in the same red as a hard error, so a script that runs fine looked like it had failed. Warnings now use the orange that the results indicator already uses for them. Fixes #2389.

The second bug is why the underline was often missing entirely. A custom-script map serializes into its link as `customNode("cMap(...)")`, and the parser gives the code inside that string argument the enclosing block — for a settings string, the anonymous multi block. The custom editor claims only the errors located in its own block, so a warning from a link-loaded script fell through to the catch-all results box: no underline, and a line:column in the message that pointed at nothing the reader could see. Editing the script reparsed it under the right block and the underline appeared, which is why this only showed on page load. `convertToMapUss` now reparses under `rootBlockIdent`, the same thing its arbitrary-script branch already did — which is why a bare `cMap(...)` in a link, the form the existing tests use, never hit this.

### DIFF
--- a/react/src/mapper/settings/map-uss.ts
+++ b/react/src/mapper/settings/map-uss.ts
@@ -51,7 +51,8 @@ function parsePreambleCustomNodeAsMapUSS(stmt: UrbanStatsASTStatement): Preamble
 
 export function convertToMapUss(uss: UrbanStatsASTStatement): MapUSS {
     if (uss.type === 'expression' && uss.value.type === 'customNode') {
-        return uss.value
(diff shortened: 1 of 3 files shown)


lukebrody:
Superseded by a branch off main; #2389's fix is already merged as #2392.

PR #2397, opened 2026-08-31
Let a USS documentation category describe itself with hre

lukebrody:
A category description on the USS documentation page can now carry code spans, subscripts and superscripts, the way the individual constants already can through `longDescription`. `getCategoryDescription` returns a `HumanReadableName` instead of a `string`, and the single call site renders it through `reifyReact`.

Nothing uses that yet. Every case still returns a plain string, which `reifyReact` passes through untouched, so no description renders differently and no reference screenshot moves. Splitting it out this way keeps the mechanical change separate from the wording it was wanted for — the string operations documentation in the branch above, where `Regex`, `lower` and a doubled backslash are all things a reader has to type exactly.

### DIFF
--- a/react/src/uss-documentation.tsx
+++ b/react/src/uss-documentation.tsx
@@ -18,6 +18,7 @@ import { DocumentationTable, renderType, USSDocumentedType } from './urban-stats
 import { constantsDocumentationData } from './uss-documentation-routing'
 import { codeStyle } from './utils/code-style'
 import { assert } from './utils/defensive'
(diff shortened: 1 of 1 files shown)


lukebrody:
Superseded: a category description is body text rendered in exactly one place, so it should just be a ReactNode. Going through HumanReadableName bought a reifyString form and unit-aware quantities that nothing consumes here, and the list element it led to forced a fabricated fallback in the link-preview card's exhaustive switch. Reopening as a smaller change.

PR #2401, opened 2026-08-31
Give every letter its own signature bit

lukebrody:
`toSignature` packed two bits per letter with `1 << ((charCode - 'a') * 2)`, but shifts in JS are 32-bit, so letters q-z wrapped around onto a-e. The signature is only a prefilter and aliasing only costs skips, so that half of the bug was invisible except as lost speed.

Bit 31 was worse. A doubled p set it, which made the signature negative, and `bitCount`'s high word term then evaluated `Math.floor(negative / 2**32) === -1` and added 32 phantom bits. Anything whose signature used bit 31 was rejected against every haystack that didn't share it, so `bitap` looking for `apple` in `aple` reported no match within 2 errors instead of 1. That is a real matching failure, not just a slow path, which is why there's a new test for it.

Signatures now give all 26 letters their own bit and spend the six spare bits on second occurrences of the letters most often repeated within a name; `bitCount` is a plain unsigned 32-bit popcount. Storage stays `Uint32Array`, so the index doesn't grow.

## Performance

Search runs about 23% faster. Timings are ms/query for 39 queries against the real index, run in headless Chrome against the dev server (median of 5 timed passes per run), with the three variants interleaved so machine drift shows up in every row rather than one:

| | round 1 | round 2 | round 3 |
| --- | --- | --- | --- |
| before | 60.51 | 60.27 | 61.31 |
| one bit per letter, no second occurrences | 56.19 | 56.79 | 56.58 |
| this PR | 46.73 | 46.87 | 47.11 |

The middle row is why the six spare bits are worth the extra code: dropping second occurrences entirely is simpler and still beats the aliasing bug, but it recovers less than a third of the win. Over the same queries, the prefilter's entry-level skip rate goes from 32.3% to 39.3% and its token-level skip rate from 65.7% to 71.3%.

Spending the spare bits on `aeiors` rather than some other six letters is worth little on its own — the six most repeated letters overall (`anoist`) land within 2% of it, which is noise here. Any reasonable choice of common letters works.


lukebrody:
!updateScreenshots

PR #2410, opened 2026-09-02
Show the mapper's computed scale bounds

### LINKED ISSUE #2000: Mapper: display calculated min/center/max when not enabled
Would be nice to show what the true values of these are to users. Example mockup:

Image

Box would be grayed out when not enabled -- user would have to enable it to be able to change the value (or perhaps trying to modify the value in the box could automatically enable it).

lukebrody:
Closes #2000. The min, center and max boxes were blank until checked, so there was no way to see what the map was using. Each unchecked one now shows that value, greyed out.

A USS function now receives the location of its own call and can record a value against a block ident through the context. Those recordings join the assignments dictionary the editors already receive, so a scale's bounds reach the box for an argument by the ident that argument's editor already derives — no new plumbing into the UI, and nothing above `scale.ts` is scale-specific.

The values come from the scale's own inverse rather than being recomputed, so they agree with the colorbar even where every value is identical and the scale spreads a range around it.

The mapper reference screenshots gain the greyed boxes and need `!updateScreenshots`.


lukebrody:
!updateScreenshots

PR #2412, opened 2026-09-02
Let a table column hold strings and booleans

### LINKED ISSUE #2408: Tables should allow boolean and string columns
Boolean columns could be rendered as ✅/❌ or sth

lukebrody:
Closes #2408. A column's values are any primitive vector rather than only numbers: booleans render as ✅/❌, strings as their text.

Such a column has no scale to place a row on, so it carries no ordinal or percentile — and its header should not promise one. That is what made `onlyColumns` per-column in the measured layout. It is derived inside `TableContents` from the rows themselves rather than passed in by each caller, so article and comparison tables keep exactly the headers they had; a column only drops the sub-headers when it has rows and none of them is a statistic.

Sorting compares strings with `localeCompare` and booleans as false < true. CSV writes `true`/`false` and the raw string with no rank columns beside them, and the link-embed card leaves a row's rank blank when the sorted column is textual.

`null` is a primitive too, so widening the argument also let `column(values=[null, null])` typecheck, where the cell fell through to the metadata branch's "See term table below". A mixed vector is already rejected as heterogenous, so nulls can only ever be a whole column of nothing — the constructor now rejects it, which keeps the three rendering paths free of null handling. That is the second commit.

Verified with `test/statistics.test.ts` under `--docker=ci --compare=true`: 59/59 assertions, every existing screenshot unchanged, and a new reference for the added test.

Stacked on #2411.


react/src/urban-stats-script/constants/table.ts
```diff
@@ -192,8 +204,20 @@ export function orderNonNan(a: number, b: number): number {
     return a - b
 }
 
+/** How a column sorts, whichever of the three kinds of value it holds. */
+export function orderCells(a: number | string | boolean, b: number | string | boolean): number {
+    if (typeof a === 'string' && typeof b === 'string') {
+        return a.localeCompare(b)
```
lukebrody: This might be a problem if the site runs in different locales. Could cause inconsistent ordering.

react/src/urban-stats-script/constants/table.ts
```diff
@@ -58,7 +67,10 @@ export const column: USSValue = {
     // eslint-disable-next-line @typescript-eslint/no-unused-vars -- needed for USSValue interface
     value: (ctx: Context, posArgs: USSRawValue[], namedArgs: Record<string, USSRawValue>, originalArgs: OriginalFunctionArgs): USSRawValue => {
         const namePassedIn = namedArgs.name as string | null
-        const values = namedArgs.values as number[]
+        const values = namedArgs.values as (number | string | boolean | null)[]
```
lukebrody: Let's link this type to a one-source-of-truth primitive type

lukebrody:
We use `number | string | boolean` and `number[] | string[] | boolean[]` let's centralize these and give them semantic meaning.

lukebrody:
Addressed the three review comments, plus two bugs found while doing it.

**Centralized value types** (`9df5c73e1`). `constants/table.ts` now owns `TableColumnValues`, `TableCellValue` derived from it as `TableColumnValues[number]`, and `TableTextValues` as `Exclude<TableColumnValues, number[]>`; the other five files use those. The raw argument cast is `USSPrimitiveRawValue[]`, which is what makes the null check beside it read as a narrowing.

Deriving the array type from the primitive instead does not work — `T extends unknown ? T[] : never` distributes `boolean` into `true | false`, so nothing of type `boolean[]` is assignable to the result. Deriving the cell type from the column type has no such trap.

**Locale-dependent sort** — agreed, and worse than two devices disagreeing: the embed card computes the same ordering inside a Worker whose locale is unrelated to the reader's, so the card and the page could disagree about one table. `orderCells` now uses a pinned `Intl.Collator('en')`, kept as a collator rather than `<`/`>` so accents and case still sort sensibly for place names. The e2e test compares against the same pinned collator.

**CSV conditions** — applied as suggested.

**Also fixed:** `Selector` collected options by calling `possibilities` once per target type into a `Set`, which dedupes by identity, so every call's fresh option objects survived — three "Custom Expression" rows here, and two on `main` today in the condition editor's right-hand side. Fixed in #2411 as `31fdca77a`, since that is the lowest branch with the problem.

Derived titles ran a column's own filter together with the table's, which #2414 fixes for tables and maps alike; merged into this branch.

Local e2e is unverified — two `--docker=ci --compare=true` runs died environmentally (the second stalled when the machine slept). The screenshots for the title, collator and selector changes will need CI's `!updateScreenshots`.

lukebrody:
!updateScreenshots

PR #2419, opened 2026-09-04
Keep the previous run's block values when execution fails

lukebrody:
The computed-bound boxes (#2410) show what the map settled on for an argument the user hasn't set, and a failed execution stops partway: with a bad `max`, `linearScale` never runs, so `min`/`center` vanish from the panel until some later run succeeds.

That also made the panel depend on which mid-edit state won the 500ms debounce in `useMapGenerator`, which is what `mapper-ux_x0`'s `syntax error in subsubfield` screenshot has been flaking on. A parse error returns the previous generator without executing at all, so the screenshot catches whichever intermediate state last executed — and whether the empty-editor state (a type error, the one that wipes the values) got a 500ms gap to itself depends on how fast testcafe dispatches keys.

Merging the two sets was the other option, but a merge can claim a variable was assigned when the failed run never assigned it, so only block values carry over. The test asserts the box still exists rather than its number: enabling `max` legitimately moves the computed `min` from 3079.46 to 3078.96, since the scale derives its bounds from each other.


lukebrody:
!updateScreenshots

PR #2421, opened 2026-09-05
Edit button on the mapper view page

### LINKED ISSUE #105: add edit button to mapmaker share page

lukebrody:
Closes https://github.com/kavigupta/urbanstats/issues/105.

The button navigates in place instead of opening a tab. The viewer is already a tab the editor opened, and sending the user forward into a third one is worse than reusing this one.

Two constraints shaped the test, neither visible in the diff. TestCafe's native automation refuses to drive a second browser window, so the test stubs `window.open` and lets the View button navigate the tab it is already in; the handler's URL construction still runs for real. The screenshot is viewport-sized because view mode positions the map absolutely, leaving the body zero pixels tall, which `screencap`'s full-page path cannot photograph.


lukebrody:
!updateScreenshots

PR #2423, opened 2026-09-05
Hide a map's missing values unless asked to show them

### LINKED ISSUE #2339: Ramp option to just disappear nans
Currently nans get a very loud color on the map, which isn't always desirable. Not sure if disappearing them should be on by default or not, but it should definitely be possible. 

You can simulate this with a condition today but like,, that's annoying

lukebrody:
Closes #2339.

Hidden means a transparent fill, not a dropped feature. The geography keeps its outline and still answers a hover or a click; only the misleading colour goes. The second e2e test clicks a real pixel inside a hidden region rather than going through `clickMapFeature`, which calls the handler directly and so would pass either way.

Two arguments rather than one because the "draw them at all" default and the "what colour" default point in opposite directions, and no single USS expression can express the latter — the furthest colour depends on the ramp, so it can only be worked out at render time.

They sit on the map constructors rather than on the ramp value, even though the issue asks for a ramp option. `RampT` is a bare `[number, string][]` shared with the generated ramp data, the `rampXyz` constants and the selector's gradient rendering; making it a record to carry two fields would ripple through all of that for no gain.

`cMapRGB` is untouched — it has a separate NaN glitch of its own, where `doRender` emits `#NaNNaNNaN`.

`#RRGGBBAA` was checked in both renderers rather than assumed: maplibre's style-spec parses it to alpha 0, and resvg renders it byte-identically to not drawing the shape. Expect screenshot churn beyond this branch's own references, since mapper settings panels gain two rows and any existing shot with missing data loses its magenta.


react/src/urban-stats-script/constants/map.ts
```diff
@@ -117,6 +129,34 @@ export const constructOutline = {
...
+        namedArgs: {
+            color: {
+                type: { type: 'concrete', value: { type: 'opaque', name: 'color' } },
+                defaultValue: createConstantExpression(null),
+            },
+        },
+        returnType: { type: 'concrete', value: missingDataType },
+    },
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars -- needed for USSValue interface
+    value: (ctx: Context, posArgs: USSRawValue[], namedArgs: Record<string, USSRawValue>, _originalArgs: OriginalFunctionArgs): USSRawValue => {
```
lukebrody: should just be able to drop _originalArgs

react/src/urban-stats-script/constants/map.ts
```diff
@@ -117,6 +129,34 @@ export const constructOutline = {
...
+        return { type: 'opaque', opaqueType: 'missingData', value: { color } }
+    },
+    documentation: {
+        humanReadableName: 'Missing Data',
+        category: 'map',
+        isDefault: true,
+        namedArgs: {
+            color: 'Color',
+        },
+        longDescription: 'Draws the geographies a map has no value for, which are otherwise left blank. Without a color, they take the one furthest from the colors the map itself uses.',
```
lukebrody: This is user facing, make it clearer and less poetic

react/unit/map-visuals.test.ts
```diff
@@ -0,0 +1,62 @@
+/* eslint-disable no-restricted-syntax -- map colours are hex, not theme colours */
```
lukebrody: How valuable is this unit test?

react/src/urban-stats-script/constants/map.ts
lukebrody: Dropped, along with the eslint-disable it needed.

react/src/urban-stats-script/constants/map.ts
lukebrody: Reworded: "Draws the geographies whose data is missing. Without this, they are not drawn at all. If no color is given, one that contrasts with the rest of the map is picked automatically."

react/unit/map-visuals.test.ts
lukebrody: Not valuable enough — deleted it. Three of its four cases only restated what the screenshots already pin. The fourth, the cluster map's extra bin, wasn't covered anywhere, so there's now a cluster-map e2e instead: a missing value that lands outside every category doesn't draw at all, which the screenshot shows as plainly as the bin numbers did.

lukebrody:
!updateScreenshots

lukebrody:
!updateScreenshots

lukebrody:
!updateScreenshots

PR #2424, opened 2026-09-05
Give an RGB map the same missing-data handling

### LINKED ISSUE #2422: cMapRGB renders NaN channels as #NaNNaNNaN
A `cMapRGB` whose `dataR`/`dataG`/`dataB`/`dataA` contain NaN — which any statistic missing for some geographies produces — emits the literal string `#NaNNaNNaN` as that geography's fill.

`clipValues` in [map.ts:398](react/src/urban-stats-script/constants/map.ts#L398) is `Math.max(0, Math.min(1, v))`, which passes NaN through untouched. [mapVisuals](react/src/mapper/map-rendering.ts#L79) then hands those channels to `doRender`, whose `hex` helper does `Math.round(NaN * 255).toString(16)` → `"NaN"`.

`cMap` doesn't have this problem: `rampColorer` routes a NaN value through [interpolateColor](react/src/utils/color.ts#L32), which substitutes the ramp's furthest color. `cMapRGB` has no equivalent path, and no ramp to derive one from, so deciding what a NaN channel should look like is part of the fix — a fixed no-data color, dropping the geography (cf. #2339), or treating a NaN channel as 0.

Noticed while working on NaN handling elsewhere in the mapper; filed separately since it is pre-existing and independent.

lukebrody:
Closes #2422. Stacked on #2423, which introduces `missingData`; review that first.

Shown missing values default to magenta on an RGB map rather than to the furthest colour `cMap` derives. An RGB map has no ramp to be furthest from, and deriving one from the thousands of colours the data itself produces would cost more per render than an opt-in, rare case is worth. Magenta is the conventional no-data colour and `constructMissingData(color=)` overrides it, so the asymmetry with `cMap` costs a reader one sentence.

The `missingData` argument moves up into the shared block in `mapConstructorArguments` so that `cMapRGB` picks it up. The order the non-RGB constructors present their arguments in is unchanged, which the reference screenshots confirm — only the two new RGB ones differ.


react/test/mapper-missing-data.test.ts
```diff
@@ -41,3 +41,25 @@ mapper(() => test)('a hidden region is still clickable', { code, geo: 'Subnation
...
+    basemap=noBasemap()
+)`
+
+mapper(() => test)('missing values on an RGB map', { code: rgbCode, geo: 'Subnational Region', universe: 'Iceland' }, async (t) => {
+    await t.expect(getErrors()).eql([])
+    await screencap(t, { removeEntireMap: false })
+
+    await checkBox(t, /^Show Missing Data/)
+    await t.expect(getErrors()).eql([])
+    await screencap(t, { removeEntireMap: false })
```
lukebrody, line 89: Add another screencap after checking the color box

react/test/mapper-missing-data.test.ts
```diff
@@ -41,3 +41,25 @@ mapper(() => test)('a hidden region is still clickable', { code, geo: 'Subnation
...
+})
```
lukebrody, line 102: Add a test case for clicking transparent region in rgb map

react/test/mapper-missing-data.test.ts
lukebrody: Added, matching the cMap test's third shot.

react/test/mapper-missing-data.test.ts
lukebrody: Added as 'a hidden region on an RGB map is still clickable', clicking the same pixel in Austurland as the cMap one.

lukebrody:
!updateScreenshots

lukebrody:
!updateScreenshots

PR #2430, opened 2026-09-05
DO NOT MERGE: thicken the histogram line so CI shows an asset diff

lukebrody:
Exists to be run, not merged. One line thickens the histogram stroke, which lands in both kinds of asset at once: `histogram-basic-article` screenshots the histogram and then saves the same SVG markup, so it appears in the diff viewer twice, once as an image and once as a text diff.

What to check on the run: the comment links to `asset-diff-viewer.html`, the text entries render a unified diff rather than a "String does not match expected value" failure, and `!updateAssets` on this PR writes both the PNGs and the gzipped XML back.

Stacked on #2429.


lukebrody:
!updateAssets

lukebrody:
!updateAssets

PR #2433, opened 2026-09-06
Group a category's pie-chart statistics under a subcategory

### LINKED ISSUE #220: Group together pie chart statistics
Right now, there's no grouping of "pie chart" statistics, that is, statistics that add to 100% (or less, if there's a missing "Other" categoryN. Doing this would make the interface much more readable

lukebrody:
Closes #220.

A subcategory is a tag on each group, not a container in the tree: a contiguous run of groups carrying the same one *is* that subcategory. The alternative, making `Category.contents` a union of groups and containers, would have rippled through `GroupIdentifier`, `allGroups`, `statParents`, `statPathToOrder`, and the eight places that just want a category's groups flat. As tagged, group ids, tree ordering, and the `?s=` link vector are untouched, so existing links and saved settings keep working. Python asserts the contiguity and id uniqueness this representation leans on.

Subcategories have their own expand toggle. The first pass gave them none, on the theory that a second level of `AnimatedCollapse` wasn't worth it, but that made Race worse in the other direction: reaching the segregation groups meant unfolding seven races. Instead of nesting the collapses in the DOM, each row names the collapse it belongs to, and a subcategory's run shows only when both it and its category are expanded.

Which runs became subcategories is a judgement call; categories that are already a single pie chart (Industry, Occupation, Generation, Religion) were left alone, since a subcategory there is a redundant row.

Checkbox grouping only — this is the structure a pie chart would need, but it draws none. The new screenshot reference was generated on arm64 and may want regenerating.

### DIFF
--- a/react/src/components/ArticleWarnings.tsx
+++ b/react/src/components/ArticleWarnings.tsx
@@ -1,7 +1,7 @@
 import React, { ReactNode } from 'react'
 
 import { useMissingGroupReasonsOfEveryGroup, useMissingGroups, useSelectedGroups } from '../page_template/statistic-settings'
-import { Category, Group, GroupIdentifier, StatPath, statPathToOrder } from '../page_template/statistic-tree'
+import { Category, Group, GroupIdentifier, StatPath, statPathToOrder, Subcategory } from '../page_template/statistic-tree'
 
 import { useScreenshotMode } from './screenshot'
 import { warningMessage } from './warning-message'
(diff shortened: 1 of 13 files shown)


lukebrody:
Reopening as a two-PR stack: the edit-tree expansion change lands first, so the subcategory work never introduces the `stat_subcategory_expanded_*` persistence that it would immediately have to delete.

PR #2434, opened 2026-09-06
Watch a --docker e2e run over VNC

lukebrody:
`chrome://inspect` shows the page but not the display around it, so native dialogs and window-sizing behaviour — exactly the class of thing that only breaks in the container — were invisible while debugging a Docker-only failure.

Two macOS quirks shaped the approach. Screen Sharing refuses VNC's "no authentication" security type and falls back to prompting for a password that doesn't exist, so `-nopw` is out and each run generates one to pass in the `vnc://` URL. And Screen Sharing has no window-level scripting — `get name of every window` errors — so closing the viewer means quitting the app. Quitting is skipped when it was already running, so a session opened by hand survives; the cost is that overlapping runs share one app and the first to finish closes it for both. Their displays are independent, at least, since the port comes from the OS.

### DIFF
--- a/react/test/scripts/run-e2e-tests-docker.ts
+++ b/react/test/scripts/run-e2e-tests-docker.ts
@@ -31,7 +31,7 @@ export async function runE2eTestsDocker(args: string[], hostArch: boolean, docke
             '--mount', 'type=tmpfs,dst=/urbanstats/react/node_modules',
             '-w', '/urbanstats/react',
             ...(['PORT', 'TESTCAFE_PORT'].flatMap(envVar => process.env[envVar] ? ['-e', `${envVar}=${process.env[envVar]}`] : [])),
-            '-e', `URBANSTATS_VNC=${vncPassword}`,
+            '-e', `URBANSTATS_VNC_PASSWORD=${vncPassword}`,
             '-e', `URBANSTATS_VNC_PORT=${vncPort}`,
             imageName,
             ...args,
--- a/react/test/scripts/run-e2e-tests-docker.ts
+++ b/react/test/scripts/run-e2e-tests-docker.ts
@@ -1,3 +1,5 @@
+import crypto from 'crypto'
+import net from 'net'
 import os from 'os'
 import path from 'path'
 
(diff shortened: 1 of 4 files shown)


react/test/scripts/run-e2e-tests.ts
```diff
@@ -64,6 +64,10 @@ if (options.headless) {
     void execa('Xvfb', [':10', '-ac'])
     process.env.DISPLAY = ':10'
     void execa('bash', ['-c', 'fluxbox >/dev/null 2>&1'])
+    if (process.env.URBANSTATS_VNC) {
+        // Retries because Xvfb above may not be listening yet
+        void execa('bash', ['-c', 'until x11vnc -display :10 -rfbport "$URBANSTATS_VNC_PORT" -passwd "$URBANSTATS_VNC" -forever -shared -quiet; do sleep 0.2; done'])
```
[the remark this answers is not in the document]
lukebrody: Applied in cf04ffaab. Confirmed the caveat first: with `-localhost`, `nc -z localhost <port>` from the Mac still connects through OrbStack's host networking, so the viewer path is unaffected.

react/test/scripts/run-e2e-tests-docker.ts
```diff
@@ -26,10 +31,64 @@ export async function runE2eTestsDocker(args: string[], hostArch: boolean, docke
             '--mount', 'type=tmpfs,dst=/urbanstats/react/node_modules',
             '-w', '/urbanstats/react',
             ...(['PORT', 'TESTCAFE_PORT'].flatMap(envVar => process.env[envVar] ? ['-e', `${envVar}=${process.env[envVar]}`] : [])),
+            '-e', `URBANSTATS_VNC=${vncPassword}`,
```
lukebrody:
```suggestion
            '-e', `URBANSTATS_VNC_PASSWORD=${vncPassword}`,
```

react/test/scripts/run-e2e-tests-docker.ts
```diff
@@ -26,10 +31,64 @@ export async function runE2eTestsDocker(args: string[], hostArch: boolean, docke
...
+// A port of its own, so parallel runs don't fight over one.
+function freePort(): Promise<number> {
+    return new Promise((resolve) => {
+        const server = net.createServer()
+        server.listen(0, () => {
+            const { port } = server.address() as net.AddressInfo
+            server.close(() => { resolve(port) })
+        })
+    })
+}
```
lukebrody, line 56: Is there a better way to do this?

react/test/scripts/run-e2e-tests-docker.ts
lukebrody: Done in c9de0df04 — the gate in `run-e2e-tests.ts` reads the same variable, so it was renamed too.

react/test/scripts/run-e2e-tests-docker.ts
lukebrody: I looked and I don't think so. Letting x11vnc pick its own port with `-autoport` is the obvious alternative, but then only the container knows the number, and getting it back out means parsing it from stdout that is already `inherit`ed to the terminal, or writing it to a file in the bind mount. Deriving one from `TESTCAFE_PORT` the way `testCafePorts` does would avoid the guesswork, but the `.envrc` convention spaces those only two apart (1337, 1339), so any fixed offset lands on a neighbouring checkout's ports.

`listen(0)` is the usual way to ask the OS for a port nobody is using. The wart is the gap between closing that socket and x11vnc binding: if something grabs the port in between, the container's `until` loop retries forever and no viewer ever appears. The tests still run, so it fails quietly rather than badly.

PR #2435, opened 2026-09-06
Derive edit-tree expansion from the selection instead of persisting it

lukebrody:
A category's expansion was a setting, so the tree opened in whatever shape the last visit left it, and a separate effect had to reopen any category a settings link was staging a change behind. Those are the same problem: what should be open is a function of what is selected right now. Deriving it when edit mode opens replaces the stored state and the effect together, and not carrying it across sessions is deliberate — a remembered expansion is only ever right by accident.

Two consequences that aren't obvious from the diff. Main now opens expanded on a default profile, so unchecking a group there no longer makes its row vanish; the tree tests that exercise the collapsed cycle collapse Main explicitly instead. And the staging e2e tests had to change target — any staged change used to be enough to prove the reopening worked, but a category with other selections is now open regardless, so they stage off a category's only selected group.

Screenshot references need regenerating.


lukebrody:
!updateScreenshots

PR #2445, opened 2026-09-07
Add a USS isNaN function

lukebrody:
USS had no way to test a number for NaN: `nanTo0` replaces it, and `x == NaN` is always false, so a script branching on missing data had to substitute a sentinel first.

Written inline rather than behind a factory, since it is the only number-to-boolean function.


lukebrody:
!updateAssets

PR #2451, opened 2026-09-07
Let a map span several universes and geography kinds

### LINKED ISSUE #1751: Allow multiple universes and types in the map maker
Eg you should be able to put in Counties in the US and census divisions in Canada on the same map

Probably the easiest way to do this is to change the universe / geography type selector to have like, a plus button that lets you do multiple, then just merge the data sources as needed.

lukebrody:
Closes #1751.

The issue suggests parallel universe/type selectors with a plus button. `MapSettings` instead carries one `geographies: {universe, geographyKind}[]`, because every consumer — the executor descriptor, the shape and centroid loaders, the row list, the inset combiner — wants the pair as a unit, and parallel arrays make a length mismatch representable but meaningless.

Merging the data sources turned out to need no merging code. The script runs once over the concatenated geographies rather than once per geography, so `population_us_census` being NaN on rows the US census does not cover is resolved by the multi-source fallback that already existed, element-wise, and `population` comes out dense across a USA + Iceland map.

Insets are the real conflict, since every universe brings a main map authored against the whole canvas. Main maps go through the comparison map's partitioner at a fill threshold of 0.6 rather than its 0.1: measured over the inset data, neighbours fill at least 0.74 (USA + Canada 1.03, France + Germany 0.76) and distant pairs at most 0.52 (USA + France 0.49, USA + Japan 0.30), so the split lands between the clusters rather than inside one. Non-main insets are then remapped onto the region their own universe covers within whichever main map it landed in; the invariant is that an inset stays over the same place on its main map, which is what the test asserts instead of recomputing the layout.

One piece is dead code on purpose: `loadOrderingDataProtobuf` returning `undefined` for a statistic a geography kind never computed. I probed all 499 statpaths across three geography kinds and none are missing — it is there so a future data build cannot turn one missing column into a blank map.


lukebrody:
!updateAssets

lukebrody:
!updateAssets

PR #2457, opened 2026-09-08
Clear a run's assignments once the next run replaces them

lukebrody:
Clearing a superseded run's values outright beats arranging for nothing to capture them. `variables` and `blockValues` are emptied as soon as the next run replaces them, which releases the generation no matter what still holds the `AssignmentsResult`.

That generation is worth releasing: a USA ZIP round trip in the mapper strands 32,967 opaque geo handles, about 12.8 MB, for the life of the page, and at the top of this stack that puts `memory_mapper_leak` over its threshold. Nothing in React is at fault. V8 gives a function scope one Context, shared by every closure created in it and holding everything any of them captures, so `Selector`'s `renderPossibility` — `useCallback`'d on deps with nothing to do with any run — shares a Context with the `setSelection` arrow that captures all of `AutoUXEditor`'s props, `assignments` among them. Which generation gets pinned depends on when those deps last changed, which is why main strands one too.

An earlier revision of this branch moved the values out of props instead, into a context holding a `Property` of stable identity. That works, but it takes fifteen files to keep a run uncapturable and still leaves main's stranded generation in place. This releases both, in five files: 69.3 MB against main's 77.1 MB, by `npm run test:memory-ab` on amd64 against a production bundle. A provider component taking the values as a prop retains as much as changing nothing; that was measured too.

The risk is that `variables` and `blockValues` are now optional and every consumer asserts them non-null. That holds only because nothing reads assignments from a run the UI has already replaced — a consumer retaining an `AssignmentsResult` across runs would find it empty rather than stale.

Bottom of a three-branch stack; replaces #2455, which a branch rename closed.


react/src/mapper/map-generator.tsx
```diff
@@ -623,12 +624,18 @@ function filterOverlaps(inset: Inset, features: GeoJSON.Feature[]): GeoJSON.Feat
 }
 
 interface MapCache {
+    assignments: Property<AssignmentsResult>
     geo?: { universe: Universe, geographyKind: typeof valid_geographies[number] } & (
         { type: 'points', centroidsByName: Map<string, ICoordinate> }
         | { type: 'polygons', polygonsByName: Map<string, GeoJSON.Geometry> }
     )
 }
```
lukebrody, line 633: Rewrite this, separate assignments from the cache

PR #2462, opened 2026-09-09
Flatten button corners in tests

lukebrody:
Antialiased rounded corners render inconsistently from run to run on the same machine, so button-heavy screenshots go flaky. #1695 fixed that for inputs and the script editor; buttons are the remaining offenders.

A root attribute plus one rule beats #1695's per-component inline style here, because button radius comes from a single element-level rule that inline styles would have to fight one call site at a time.

Quiz answer cards keep their radius: they're `role=button` divs, and flattening them moves the quiz screenshots much more than this is worth.


lukebrody:
!updateAssets

lukebrody:
!updateAssets

PR #2468, opened 2026-09-12
Take a map's unit from what the script states, not from its data's documentation

kavigupta:
Stacked on #2467, which this depends on: it removes a fall back that only worked for the mapper, so both callers need the shared one from that PR first.

`computeCommonMap` recorded a unit for the map from one of two places:

```ts
if (unitArg) {
    unit = unitArg.value.unit as UnitType          // what the script stated
}
else {
    const inferredUnit = originalArgs.namedArgs.data.documentation?.unit   // a guess
    ...
}
```

The second is a guess taken from whatever statistic the data argument evaluated to. It only works when the data is a bare statistic. For an expression such as `population / area` there is no single statistic to ask, so it produced nothing at all, and the ramp lost its unit.

The map's unit now means only what the script stated. Both callers already fall back to working the unit out from the script, which handles bare statistics and expressions alike:

| script | before | after |
| --- | --- | --- |
| `cMap(data=high_temp)` | `°F`, from the statistic's documentation | `°F`, worked out from the script |
| `cMap(data=population / area)` | no unit | `/km²` |
| `cMap(data=high_temp, unit=unitNumber)` | `unitNumber`, as stated | unchanged |

So nothing loses a unit, and computed expressions gain one.

`computeCommonMap` no longer takes `originalArgs`, since reading the data argument's documentation was the only thing it used them for.

https://claude.ai/code/session_01KHKp7QCPqi74raCmowqSZH


kavigupta:
Folded into #2467. The two changes have to travel together: this one removes a fall back that only the mapper had, so the shared one from #2467 must exist first, and keeping them apart left #2467 with a parameter that could not be named accurately.

PR #2470, opened 2026-09-12
Put the embed card's tile credit under the map it credits

lukebrody:
In the card footer, across the page from the map, the credit read as a credit for the whole card. The map panel gives up the height to carry it instead.

The map card keeps its footer credit because there the footer already abuts the map.

`embed_worker.test.ts` card screenshots need regenerating.


lukebrody:
!updateAssets

PR #2473, opened 2026-09-12
Give the mapper the same credit treatment

lukebrody:
Top of the stack, and the end of the attribution-control subclass, whose only job was making the credit start expanded rather than as a collapsed "i".

The credit hangs off the whole layout rather than the main inset. Attaching it to whichever inset is `mainMap` was arbitrary — it credits all of them.

Mapper screenshot references need regenerating.


lukebrody:
!updateAssets

PR #2499, opened 2026-09-14
Clicking a mapper geography goes to its own universe

lukebrody:
Clicking a French region on a USA+France map opened its article in the article's default universe rather than France. A multi-geography map has no single universe, so `EditMapperPanel` provides no universe context at all, and every click fell through to `defaultArticleUniverse`.

The universe is carried on the feature rather than passed to the map component because the click handler — and the TestUtils path the e2e tests drive — only has a longname to work from, so a per-feature lookup keeps both going through the same resolution.

Where two selected universes contain the same longname, the later-listed one wins. That is not a considered choice so much as whose geometry the existing merge already keeps; ranking universes by specificity looked like more machinery than the case deserves.

Exported GeoJSON now carries a `universe` per feature, which is a side effect rather than a goal.

The new e2e case fails on `main` with `expected 'http://localhost:8000/article.html?longname=%C3%8Ele-de-France%2C+France&s=...' to include 'universe=France'`.


lukebrody:
!updateSnapshots

